PR #15, opened 2022-10-13
Update README.md

dllu:
Correct a factual inaccuracy regarding Livox lidars.

Although they are rectangular instead of round, they are still mechanical since they scan using Risley prisms.


nachovizzo:
@dllu Thanks for letting us know. Honestly we haven't really dive deep into the scanning technology, for us, it's all about point clouds, and everyone else is calling this "solid-state". So, thanks for clarifying this!!!

PR #26, opened 2023-01-11
Remove deprecated use of np.str

### LINKED ISSUE #25: AttributeError: module 'numpy' has no attribute 'str'
First of all, thank you so much for publishing the source code!

Starting with numpy version 1.20.0, the use of `np.str` has been [deprecated](https://numpy.org/doc/1.20/release/1.20.0-notes.html#deprecations) and its support has been [completely removed](https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations) in numpy version 1.24.0, leading to an `AttributeError: module 'numpy' has no attribute 'str'` when trying to use KISS-ICP for example with PCD files.

The [user manual suggests](https://numpy.org/doc/1.20/release/1.20.0-notes.html#deprecations) to instead use the plain Python `str`:

> To give a clear guideline for the vast majority of cases, for the types bool, object, str (and unicode) using the plain version is shorter and clear, and generally a good replacement.

markuspi:
* Replace `np.str` with `str`

Fixes #25

### DIFF
--- a/src/kiss_icp/datasets/generic.py
+++ b/src/kiss_icp/datasets/generic.py
@@ -43,7 +43,7 @@ class GenericDataset:
                     if any(fn.endswith(ext) for ext in supported_file_extensions())
                 ]
             ),
-            dtype=np.str,
+            dtype=str,
         )
         if len(self.scan_files) == 0:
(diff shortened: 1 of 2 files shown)


nachovizzo:
Thanks! My linter was always yelling at me, but I never paid attention.

PR #41, opened 2023-01-17
Adding "time" option to timestamps msg fields in rosbag.py

RoniEmad:
Velodyne lidar ros driver outputs pointcloud msg with timestamps field name "time", so I added it as an option.

### DIFF
--- a/src/python/kiss_icp/datasets/rosbag.py
+++ b/src/python/kiss_icp/datasets/rosbag.py
@@ -69,12 +69,13 @@ class RosbagDataset:
 
         t_field = None
         for field in msg.fields:
-            if field.name == "t" or field.name == "timestamp":
+            if field.name == "t" or field.name == "timestamp" or field.name == "time":
                 t_field = field.name
         timestamps = np.ones(points.shape[0])
         if t_field:
             timestamps = np.array(list(self.pc2.read_points(msg, field_names=t_field)))
-            timestamps = timestamps / np.max(timestamps)
+            if t_field != "time":
+                timestamps = timestamps / np.max(timestamps)
         return points.astype(np.float64), timestamps
 
     def check_for_topics(self):
--- a/src/python/kiss_icp/datasets/rosbag.py
+++ b/src/python/kiss_icp/datasets/rosbag.py
@@ -69,13 +69,12 @@ class RosbagDataset:
 
         t_field = None
         for field in msg.fields:
-            if field.name == "t" or field.name == "timestamp" or field.name == "time":
+            if field.name in ["t", "timestamp", "time"]:
                 t_field = field.name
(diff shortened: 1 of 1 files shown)


src/python/kiss_icp/datasets/rosbag.py
```diff
@@ -69,12 +69,13 @@ def read_point_cloud(self, bagfile: Path, topic: str, idx: int):
 
         t_field = None
         for field in msg.fields:
-            if field.name == "t" or field.name == "timestamp":
+            if field.name == "t" or field.name == "timestamp" or field.name == "time":
```
nachovizzo, line 72: Would you mind improving this a bit? :
```py
if field.name in ["t", "timestamp", "time"]:
```

src/python/kiss_icp/datasets/rosbag.py
```diff
@@ -69,12 +69,13 @@ def read_point_cloud(self, bagfile: Path, topic: str, idx: int):
...
                 t_field = field.name
         timestamps = np.ones(points.shape[0])
         if t_field:
             timestamps = np.array(list(self.pc2.read_points(msg, field_names=t_field)))
-            timestamps = timestamps / np.max(timestamps)
```
nachovizzo, line 77: I'm more into one-liners for these things, but it's personal taste
```py
timestamps = timestamps / np.max(timestamps) if t_field != "time" else timestamps
```

nachovizzo (changes requested):
Thanks for the contributions!

src/python/kiss_icp/datasets/rosbag.py
RoniEmad: Done c297133565d7aeb557b1a9d549ded93747722ef1

src/python/kiss_icp/datasets/rosbag.py
RoniEmad: Done c297133565d7aeb557b1a9d549ded93747722ef1

nachovizzo:
Amazing, thanks @RoniEmad .

FYI: We are currently working on the ROS wrapper ;)

RoniEmad:
Amazing Work.
I was working on the ROS wrapper too :)

nachovizzo:
Great, do you have an open source implementation so we can merge efforts maybe?

I'm pushing the last bis of the python API back to C++ so it's easier to wrap

RoniEmad:
This is my ros wrapper
https://github.com/RoniEmad/kiss-icp/tree/ros_wrapper

It's not complete yet, and needs to be restructured and cleaned.

To use ros, we need to pass the argument --dataloader ros and --topic

It subscribes to the topic and iterate through the pipeline once every call back.
Odometry publisher is not implemented yet.

PR #44, opened 2023-01-17
Visualizer getting slow with big number of poses

RoniEmad:
When using the visualizer, I found that the pipeline FPS is getting slower as the number of frames increases.

It starts with 25FPS and by the time it goes under 2FPS.

The issue is in **_o3d.geometry.TriangleMesh.create_sphere()_** which is not efficient, so the visualization gets slower as the number of spheres(trajectory poses) increases.

https://github.com/PRBonn/kiss-icp/blob/d848ab891ea6684d41401cc000a6e98eb7f7a3a0/src/python/kiss_icp/tools/visualizer.py#L224-L232

Using rosbags, it takes too long to process a bag when viewing the trajectory.

I am working on a ROS wrapper, this was causing a problem that when the delay increases the pipeline starts to miss messages and the odometry gets crazy.

Using **_o3d.geometry.PointCloud()_** instead solved the issue.
However there a small issue is that all pointclouds in the same open3d window must have the same point size using **_viz.update_geometry()_**.

### DIFF
--- a/src/python/kiss_icp/datasets/rosbag.py
+++ b/src/python/kiss_icp/datasets/rosbag.py
@@ -69,12 +69,13 @@ class RosbagDataset:
 
         t_field = None
         for field in msg.fields:
-            if field.name == "t" or field.name == "timestamp":
+            if field.name == "t" or field.name == "timestamp" or field.name == "time":
                 t_field = field.name
         timestamps = np.ones(points.shape[0])
         if t_field:
             timestamps = np.array(list(self.pc2.read_points(msg, field_names=t_field)))
-            timestamps = timestamps / np.max(timestamps)
+            if t_field != "time":
+                timestamps = timestamps / np.max(timestamps)
         return points.astype(np.float64), timestamps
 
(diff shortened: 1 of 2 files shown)


nachovizzo:
@RoniEmad Have you tried disabling the trajectory? You can just press `T` on the keyboard and this should solve the problem.

The trajectory path is not Ideal at all, but should only used for debugging

RoniEmad:
Yes, it works well when I press T.
But in case I need to see the trajectory path for debugging in real-time, I can't because of the delay.

nachovizzo:
I see, I guess this won't be a big problem once the ROS wrapper is ou there. I'll skip integrating this since it's suboptimal the problem of the point size, the trajectory in the visualizer is a nice to have feature...but it's to complicated to implement properly on my humble opinion


I'll keep this in mind when using rviz ;) there is much easier to debug this type of scenarios

RoniEmad:
Great, Thank you for your reply :)

PR #45, opened 2023-01-18
dataloader bug fixes and optimization for kiss icp

sumanthrao1997:


nachovizzo:
Amazing! Thanks a lot for your contribution. And I didn't know that the DL had a bug! Thanks for spotting this.

If it's not too much to ask, here are some tasks I'd like to solve before merging this

- [ ] `black .` # format the code
- [ ] Add a method `self.apply_calibration` that brings the poses estimated by KISS-ICP to the same coordinate frame of the GT poses
- [ ] Remove the `pcd.transfrom`
- [ ]

PR #50, opened 2023-01-25
Improve handling of results directory

markuspi:
- Generate name of results directory only once during construction of OdometryPipeline to ensure consistency.
  This prevents the result files being spread over multiple directories when the writing process happens to be at the turnover of a full second
- Fix error that happens when the `latest` symlink is broken.
  Previously, the old symlink would not be removed when it was in a broken state due to the target directory no longer existing. This resulted in an error during `os.symlink`


nachovizzo:
Thanks for fixing this, it was indeed a bit annoying !

Issue #56, opened 2023-02-03
Why slow in my own dataset?


qpc001:
Using VLP-16 .

qpc001:
` rostopic hz /odometry_node/odometry -w 10
subscribed to [/odometry_node/odometry]
average rate: 5.502
	min: 0.158s max: 0.219s std dev: 0.02125s window: 6
average rate: 5.679
	min: 0.141s max: 0.219s std dev: 0.02604s window: 10
average rate: 6.262
	min: 0.141s max: 0.177s std dev: 0.01448s window: 10
average rate: 6.215
	min: 0.146s max: 0.188s std dev: 0.01448s window: 10
average rate: 5.644
	min: 0.151s max: 0.221s std dev: 0.02164s window: 10
average rate: 5.857
	min: 0.154s max: 0.186s std dev: 0.00941s window: 10
average rate: 5.184
	min: 0.160s max: 0.231s std dev: 0.02504s window: 10
average rate: 4.467
	min: 0.165s max: 0.316s std dev: 0.04075s window: 10
average rate: 4.124
	min: 0.158s max: 0.321s std dev: 0.05234s window: 10
average rate: 4.384
	min: 0.158s max: 0.321s std dev: 0.04411s window: 10
`

qpc001:
rostopic hz /odometry_node/odometry -w 10
subscribed to [/odometry_node/odometry]
average rate: 5.502
	min: 0.158s max: 0.219s std dev: 0.02125s window: 6
average rate: 5.679
	min: 0.141s max: 0.219s std dev: 0.02604s window: 10
average rate: 6.262
	min: 0.141s max: 0.177s std dev: 0.01448s window: 10
average rate: 6.215
	min: 0.146s max: 0.188s std dev: 0.01448s window: 10
average rate: 5.644
	min: 0.151s max: 0.221s std dev: 0.02164s window: 10
average rate: 5.857
	min: 0.154s max: 0.186s std dev: 0.00941s window: 10
average rate: 5.184
	min: 0.160s max: 0.231s std dev: 0.02504s window: 10
average rate: 4.467
	min: 0.165s max: 0.316s std dev: 0.04075s window: 10

PR #57, opened 2023-02-03
<fix>(kiss_icp_ros/CMakeLists.txt): adding BUILD_TYPE -> RELEASE.

qpc001:
fixing the issue #56


nachovizzo:
Does this fix the problem? The ROS wrapper was not built with optimization flags, but all the computation is happening in the kiss_icp_core lib: 

https://github.com/PRBonn/kiss-icp/blob/e269bd12eb25bc65c424ed01e16af5fde463c118/src/cpp/kiss_icp/CMakeLists.txt#L31

In any case it doesn't hurt to add this, so merging it ;)

Thanks for the contribution!

PR #61, opened 2023-02-10
Treat optimized rotation as so3 rather than RPY

mattalvarado:
It appears the current optimization routine treats the optimized rotation component of the transformation as RPY. I believe this instead should be treated as a lie algebra of SO3. For small update rotations the differences are negligible, but I believe this update is still warranted for correctness.

See section 2.4 of this reference (https://www.ethaneade.org/lie.pdf) for details on the derivation of the `J_r` parameter in `ComputeUpdate`, and the representation of the underlying rotation.

### DIFF
--- a/src/cpp/kiss_icp/core/Registration.cpp
+++ b/src/cpp/kiss_icp/core/Registration.cpp
@@ -62,10 +62,11 @@ struct ResultTuple {
 
 Eigen::Matrix4d TransformVector6dToMatrix4d(const Eigen::Vector6d &x) {
     Eigen::Matrix4d output = Eigen::Matrix4d::Identity();
-    output.block<3, 3>(0, 0) = (Eigen::AngleAxisd(x(2), Eigen::Vector3d::UnitZ()) *
-                                Eigen::AngleAxisd(x(1), Eigen::Vector3d::UnitY()) *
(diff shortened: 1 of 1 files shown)


nachovizzo:
@mattalvarado Thanks for your contribution. There was a small bug (if you test the results you will see) But @tizianoGuadagnino now converted everything to Eigen::Isometry3d which should be easier to read, follow, etc.

We also merged (and reverted) your change, as a way to acknowledge your contribution. Thanks for real

nachovizzo:
@mattalvarado We just moved to Sophus and I think you would like it ;)

Please feel free to comment on this change
https://github.com/PRBonn/kiss-icp/blob/d28a3cb7acec8d523803d70e339bc50c8c92c078/cpp/kiss_icp/core/Registration.cpp#L97

mattalvarado:
> @mattalvarado We just moved to Sophus and I think you would like it ;)
> 
> Please feel free to comment on this change
> 
> https://github.com/PRBonn/kiss-icp/blob/d28a3cb7acec8d523803d70e339bc50c8c92c078/cpp/kiss_icp/core/Registration.cpp#L97

Looks great! I am a big fan of this change! Personally I typically use [manif](https://github.com/artivis/manif), but Sophus seems like it has nearly identical syntax.

PR #63, opened 2023-02-10
Improve KISSConfig

markuspi:
## Problem

When specifying a custom YAML config, all config values needed to be specified. There was no fallback behavior to selectively override values.

## Solution

This PR introduces `pydantic` to validate and manage the configration.
The default values are defined in `config.py`.
YAML config files can override as many or as few items as needed.
Cleaned up generated yaml file: Output file is plain YAML without `!!python/object/new:easydict.EasyDict`

## Changes

* Remove `easydict` dependency
* Add `pydantic` dependency
* Change dataclass-based config to pydantic based config
* Add custom loader for YAML
* Remove default yaml path from CLI. Default values are now retrieved directly from config.py.
* Rename `default.yaml` to `basic.yaml` to avoid confusion. Editing this file only has effect when it is referenced via `--config`!


src/python/kiss_icp/pipeline.py
```diff
@@ -42,9 +41,7 @@ class OdometryPipeline:
     def __init__(
         self,
         dataset,
-        config: Path,
-        deskew: Optional[bool] = False,
-        max_range: Optional[float] = None,
+        config: KISSConfig,
```
nachovizzo, line 44: This change is in general desire, but I see the disadvantage of always having to build a `KISSConfig` object to use the pipeline. The other arguments are more "standard" which basically means I can fire up a python shell now, and `from kiss_icp.pipeline import Odometrypipeline` and run it as it it.

Would you mind reverting this back to the old bheavior?

src/python/kiss_icp/tools/cmd.py
```diff
@@ -93,8 +93,8 @@ def kiss_icp_pipeline(
         callback=name_callback,
         help="[Optional] Use a specific dataloader from those supported by KISS-ICP",
     ),
-    config: Path = typer.Option(
-        f"{get_package_dir()}/config/default.yaml",
```
nachovizzo, line 97: `get_package_dir` function can also be removed

nachovizzo (changes requested):
I think I end up requesting a lot. Please tell me if you can't solve all the issues, I will merge your changes and modify it afterwards.

Thanks again for your contributions!!!

setup.py
```diff
@@ -68,10 +68,10 @@ def cmd_exists(cmd):
         package_data={"kiss_icp": ["config/default.yaml"]},
         install_requires=[
             "PyYAML",
-            "easydict",
             "natsort",
             "numpy",
             "plyfile",
+            "pydantic",
```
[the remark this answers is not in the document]
nachovizzo: This is possible, but it's a pain. So, dropping this idea

nachovizzo:
Merging these changes in #69

markuspi:
Thanks for refining 👍

PR #71, opened 2023-02-17
Add ROS2 support

shrijitsingh99:
I tested on some bag files I had and everything seemed to work. 
There is one issue where it `ros2` command is not able to the `kiss_icp` package but if I run the binaries directly it works. THis may be an issue with my PC only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,3 +44,4 @@ find_package(catkin QUIET)
 if(BUILD_ROS_WRAPPER OR ament_cmake_FOUND OR catkin_FOUND)
   add_subdirectory(src/cpp/kiss_icp_ros)
 endif()
+
--- a/package.xml
+++ b/package.xml
@@ -29,7 +29,11 @@
   <description>KISS-ICP</description>
   <maintainer email="ignaciovizzo@gmail.com">ivizzo</maintainer>
   <license>MIT</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
   <export>
-    <build_type>cmake</build_type>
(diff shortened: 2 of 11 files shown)


nachovizzo:
Hello @shrijitsingh99 thanks for opening a PR! Just out of curiosity, have you checked this fork: https://github.com/HIPERT-SRL/kiss-icp? It looks like they went a bit deeper and also provided launch files in python, etc. 

Would you mind trying to take inspiration from that branch? It looks like this PR is more like a "draft". What do you think? 

Thanks again!

shrijitsingh99:
@nachovizzo I wasn't aware of that repo when I worked on this. Both implementations are pretty similar, though one difference is that I inherited from the `Node` class for `OdometeryServer`, this allows for composing the node if needed.

Sorry, about putting an incomplete PR. I had added some launch file locally but never got to pushing them.
I added everything now, so this PR should be ready after some testing.

nachovizzo:
> @nachovizzo I wasn't aware of that repo when I worked on this. Both implementations are pretty similar, though one difference is that I inherited from the `Node` class for `OdometeryServer`, this allows for composing the node if needed.
> 
> Sorry, about putting an incomplete PR. I had added some launch file locally but never got to pushing them. I added everything now, so this PR should be ready after some testing.

Please! No need to say sorry!!! Thanks for contributing!

I'm now integrating these changes into our local dev pipeline. I will need to make some changes, if you can give me write permissions to your branch I can directly push there. If not, I can fork your fork, so your contribution is still clear.

nachovizzo:
Update: started to work on nacho/tweak_ros2_node

PR #75, opened 2023-02-22
Improve rosbag2 dataloader performace

markuspi:
* This PR copies the much more efficient numpy implementation of `read_points` from [common_interfaces/sensor_msgs_py](https://github.com/ros2/common_interfaces/blob/4bac182a0a582b5e6b784d9fa9f0dabc1aca4d35/sensor_msgs_py/sensor_msgs_py/point_cloud2.py), which was originally contributed by @Flova via https://github.com/ros2/common_interfaces/pull/175

* Even though `sensor_msgs_py` is available as a ROS2 package, the optimized code was not backported and thus is not available e.g. in the ros-galactic repository, so copying the code seems the best approach

* I copied the methods mostly as-is, see e1bcabe21dbfbb92dfab8af8317e5703908b39fc for the necessary tweaks.

* This PR results in a 300 to 400% speed up when dealing with ros2 bags


markuspi:
Looks like I missed that the ros1 dataloader also uses the same method... @nachovizzo what's your preferred way of resolving this, since you use the `rosbag` module for ros1 instead of `rosbags`?

nachovizzo:
@markuspi Amazing work, I'm merging these and just changing some minor formatting things. I've tested it and it works amazingly faster. So after fixing the format, I will launch a patch-release, so it's available for the `pip install kiss-icp` package.

Great work! Thanks a lot for contributing!

PR #85, opened 2023-02-27
Nacho/add gh actions

nachovizzo:
Testing Github actions infrastructure.

NOTE: If you are reading this and know how to do it, please feel free to open a PR

- [x] Fix sophus windows build, related to https://github.com/strasdat/Sophus/issues/394. Done: https://github.com/strasdat/Sophus/pull/502
- [x] Add pypi token to automatically deploy python packages

## Gitlab-ci

This is how a full gitlab ci pipeline looks now, we should make sure that everything is also captured in the github actions
image


ahans:
How about using [pre-commit](https://pre-commit.com/) for clang-format and black? There's a GH action available that makes integration fairly trivial.

nachovizzo:
> How about using [pre-commit](https://pre-commit.com/) for clang-format and black? There's a GH action available that makes integration fairly trivial.

Would you like to contribute it :) I'm 100% new to GH actions, and I'm honestly kind of lost

nachovizzo:
> How about using [pre-commit](https://pre-commit.com/) for clang-format and black? There's a GH action available that makes integration fairly trivial.

Adding pre-commit in #102 , not the biggest fan, but if someone finds this helpful, then, why not

ahans:
> > How about using [pre-commit](https://pre-commit.com/) for clang-format and black? There's a GH action available that makes integration fairly trivial.
> 
> Adding pre-commit in #102 , not the biggest fan, but if someone finds this helpful, then, why not

What I meant was using this as a framework for all kinds of style checks. You don't have to use it as an actual git hook. I do that in none of the projects I use pre-commit in, because I like to have some more control. What it does give you is central configuration of style checks/tools and a one-stop tool you can also use locally to check and fix stuff. That you can then also do that in CI via a GH action is a nice bonus. When the check fails, it even prints a message along the lines of: "If you see this in CI, run this command to fix it." So there's not even a need for documentating this.

I would have loved to put up a PR for this, but I would have needed a bit more time. I can still do it to make it the way I meant with my original comment if you want to.

Regarding the other PR, I will finish that up later today. Noticed that you pulled in main. Please wait before merging this. Let's do things proper. Thank you!

PR #92, opened 2023-03-08
Add options to toggle odom and alias tf

willcbaker:
This feature introduces parameters to toggle on/off publishing of TFs to avoid tf conflicts from other sources.

For examples:
- if this were used as an odometry signal in http://wiki.ros.org/robot_localization you need to toggle off the odom->base_link tf
- if not using base link frame_id but the tf already exists (e.g. from robot_state_publisher) do not publish an identity transform

### DIFF
--- a/ros/ros1/OdometryServer.cpp
+++ b/ros/ros1/OdometryServer.cpp
@@ -46,6 +46,7 @@ namespace kiss_icp_ros {
 
 OdometryServer::OdometryServer(const ros::NodeHandle &nh, const ros::NodeHandle &pnh)
     : nh_(nh), pnh_(pnh) {
+    bool alias_transform;
     pnh_.param("child_frame", child_frame_, child_frame_);
     pnh_.param("odom_frame", odom_frame_, odom_frame_);
     pnh_.param("max_range", config_.max_range, config_.max_range);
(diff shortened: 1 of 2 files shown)


nachovizzo:
Hi @willcbaker sorry for the delay, and thanks for the contribution!

This is indeed a problem, to make this PR perfect I'd ask some stuff:

- [ ] Add the parameter to the launch file, so it's easier to toggle on/off this parameter
- [ ] [Optionally] Implement for ROS2. If you have no time, I'd do it anyway after merging this change.

ros/ros1/OdometryServer.cpp
```diff
@@ -46,6 +46,7 @@ namespace kiss_icp_ros {
 
 OdometryServer::OdometryServer(const ros::NodeHandle &nh, const ros::NodeHandle &pnh)
     : nh_(nh), pnh_(pnh) {
+    bool alias_transform;
```
nachovizzo, line 49: Do we even need this duplicate variable? Why not using `publish_odom_tf_` instead?

nachovizzo:
Closing this as I haven't received any response

ros/ros1/OdometryServer.cpp
willcbaker: these are 2 different options - 
turning off alias transforms disables publishing the base_link to sensor frames
which is independent from publishing the odom message

e.g.
some users might want odom published but already have a base_link -> sensor frame defined in URDF
other users might not have tf defined (and need this published, arguable this is a strange behavior in general for ROS in my opinion) but _do_ want the odom frame published

PR #100, opened 2023-03-24
Enable and fix warnings

ahans:
This enables the commonly used warning flags for msvc/gcc/clang and fixes the warnings reported.


cpp/kiss_icp/metrics/Metrics.cpp
```diff
@@ -148,8 +148,8 @@ std::tuple<float, float> SeqError(const std::vector<Eigen::Matrix4d> &poses_gt,
         r_err += it.r_err;
     }
 
-    double avg_trans_error = avg_trans_error = 100.0 * (t_err / float(err.size()));
-    double avg_rot_error = avg_rot_error = 100.0 * (r_err / float(err.size())) / 3.14 * 180.0;
+    double avg_trans_error = 100.0 * (t_err / static_cast<double>(err.size()));
+    double avg_rot_error = 100.0 * (r_err / static_cast<double>(err.size())) / 3.14 * 180.0;
```
ahans, line 152: The changes to the cast of `err.size()` weren't strictly necessary, but 1) casting to `float` when everything else is `double` didn't seem like the right thing and 2) doing `float(...)` or `double(...)` is essentially a C-style cast, which in C++ should be avoided.

cpp/kiss_icp/pipeline/KissICP.hpp
```diff
@@ -80,8 +77,8 @@ class KissICP {
     // KISS-ICP pipeline modules
     std::vector<Sophus::SE3d> poses_;
     KISSConfig config_;
-    AdaptiveThreshold adaptive_threshold_;
     VoxelHashMap local_map_;
+    AdaptiveThreshold adaptive_threshold_;
```
ahans, line 81: gcc complained about a mismatch between the order here and the initialization lists (`-Wreorder`).

cpp/kiss_icp/pipeline/KissICP.hpp
```diff
@@ -57,10 +57,7 @@ class KissICP {
           local_map_(config.voxel_size, config.max_range, config.max_points_per_voxel),
           adaptive_threshold_(config.initial_threshold, config.min_motion_th, config.max_range) {}
 
-    KissICP()
-        : local_map_(config_.voxel_size, config_.max_range, config_.max_points_per_voxel),
-          adaptive_threshold_(config_.initial_threshold, config_.min_motion_th, config_.max_range) {
-    }
+    KissICP() : KissICP(KISSConfig{}) {}
```
ahans, line 60: Also not strictly necessary to fix the warnings, but seemed like a good chance to reduce redundancy and more clearly state that we're using the `KISSConfig` defaults.

nachovizzo:
Wow thanks a lot! Some comments, and ready to merge!

cpp/kiss_icp/metrics/Metrics.cpp
nachovizzo: True, this code was mainly not touched as was copy-psted from the kitti-dev benchmark code. Maybe some metrics will change, but we are not submitting to the benchmark again, so it looks like a safe move ;)

cpp/kiss_icp/pipeline/KissICP.hpp
nachovizzo: Yes! amazing! we love to chope down code

cpp/kiss_icp/pipeline/KissICP.hpp
nachovizzo: lgtm

cpp/kiss_icp/CMakeLists.txt
```diff
@@ -33,6 +33,10 @@ set(CMAKE_BUILD_TYPE Release)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
nachovizzo, line 36: What about using generator expressions? I'm just asking, not requiring to change

```cmake
  add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Wall>")
  add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Wextra>")
  add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wall>")
  add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wextra>")
  add_compile_options("$<$<COMPILE_LANG_AND_ID:CXX,GNU>:-pedantic>")
```

nachovizzo: I'd also add windows here: ` $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/W4 /WX >`

ahans: I noticed you have a function ` set_global_target_properties`  to set flags. I understand that this is the "modern" CMake way to do it. Since there's always the risk of forgetting to call this, I'd prefer to set stuff like that globally for the entire project. But in this case, where you have it already, `CompilerOptions.cmake`  is definitely the place for this, I moved it there. Also added `MSVC`.

cpp/kiss_icp/cmake/CompilerOptions.cmake
```diff
@@ -22,8 +22,14 @@
 # SOFTWARE.
 function(set_global_target_properties target)
   target_compile_features(${target} PUBLIC cxx_std_17)
-  target_compile_options(${target} PRIVATE $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-fcolor-diagnostics>)
-  target_compile_options(${target} PRIVATE $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-fdiagnostics-color=always>)
+
+  set(GCC_WARNINGS_FLAGS -Wall -Wextra -Wpedantic -Werror)
```
ahans, line 26: Since you suggested ` /WX` for the MSVC build, I think it's only appropriate to use `-Werror`  with the rest. Most folks find this too extreme, it can also be a bit annoying during development. But I'm generally all for it. Dropping both `-Werror`  and `/WX`  would also be an option, however.

cpp/kiss_icp/core/Registration.cpp
```diff
@@ -83,10 +83,10 @@ Sophus::SE3d AlignClouds(const std::vector<Eigen::Vector3d> &source,
             auto Weight = [&](double residual2) { return square(th) / square(th + residual2); };
             auto &[JTJ_private, JTr_private] = J;
             for (auto i = r.begin(); i < r.end(); ++i) {
-                const auto &[J_r, r] = compute_jacobian_and_residual(i);
-                const double w = Weight(r.squaredNorm());
+                const auto &[J_r, residual] = compute_jacobian_and_residual(i);
```
ahans, line 86: Shadowing the `r` parameter of the lambda here is something that MSVC complained about.

cpp/kiss_icp/3rdparty/tsl_robin/tsl_robin.cmake
```diff
@@ -28,7 +28,7 @@ if(NOT tessil_POPULATED)
   FetchContent_Populate(tessil)
   add_library(robin_map INTERFACE)
   add_library(tsl::robin_map ALIAS robin_map)
-  target_include_directories(robin_map INTERFACE "$<BUILD_INTERFACE:${tessil_SOURCE_DIR}/include>")
+  target_include_directories(robin_map SYSTEM INTERFACE "$<BUILD_INTERFACE:${tessil_SOURCE_DIR}/include>")
```
ahans, line 31: This was necessary to silence some MSVC warnings about includes from this lib. I also checked the other 3rd party libs. There you already do it, so probably just an oversight here.

cpp/kiss_icp/CMakeLists.txt
nachovizzo: Great, I also agree on that globally might look safer on the surface.

But the key idea behind "Modern cmake", something I like to call "target oriented cmake" is that each target is well defined without contaminating other ones

As a real world example, if you would like to include this project in a (for example) ROS context, and you would 'add_subdirectory(kiss_icp)', then nothing else would compile. Since now the 'Werror" would be propagated to all targets in the build system

Might seem like a corner case, but believe, there is nothing more annoying than working with projects like this 😉

cpp/kiss_icp/cmake/CompilerOptions.cmake
nachovizzo: Yes you are right, is too annoying. But I feel like the code won't change much. We only must ensure that these are private options and don't get propohated to dependent targets

cpp/kiss_icp/core/Registration.cpp
nachovizzo: Damn 🤔 why is this even working? I'll need to look further into this

cpp/kiss_icp/3rdparty/tsl_robin/tsl_robin.cmake
nachovizzo: Make sense, thanks! 🤟

nachovizzo:
@ahans the windows build is failing, probably cuz MSVC is too annoyting with warnings, don't worry. I will take over this one ;)

cpp/kiss_icp/cmake/CompilerOptions.cmake
```diff
@@ -22,9 +22,30 @@
...
   target_compile_features(${target} PUBLIC cxx_std_17)
-  target_compile_options(${target} PRIVATE $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-fcolor-diagnostics>)
-  target_compile_options(${target} PRIVATE $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-fdiagnostics-color=always>)
-
+  target_compile_options(
+    ${target}
+    PRIVATE # MSVC
+            $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/W4>
+            $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/WX>
+            $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/external:W0> # disable -isystem header warnings
```
nachovizzo, line 30: https://learn.microsoft.com/en-us/cpp/build/reference/external-external-headers-diagnostics?view=msvc-170

ahans:
> @ahans the windows build is failing, probably cuz MSVC is too annoyting with warnings, don't worry. I will take over this one ;)

Interesting, it was working for me. But I was using a newer version, maybe they changed something there.

nachovizzo:
> > @ahans the windows build is failing, probably cuz MSVC is too annoyting with warnings, don't worry. I will take over this one ;)
> 
> Interesting, it was working for me. But I was using a newer version, maybe they changed something there.

Which version? The problem seems to be that is not detecting the system headers as headers :thinking:

ahans:
> > > @ahans the windows build is failing, probably cuz MSVC is too annoyting with warnings, don't worry. I will take over this one ;)
> > 
> > 
> > Interesting, it was working for me. But I was using a newer version, maybe they changed something there.
> 
> Which version? The problem seems to be that is not detecting the system headers as headers thinking

It was the latest, which is probably some 2022 one. But it's very possible I missed something.

nachovizzo:
> > > > @ahans the windows build is failing, probably cuz MSVC is too annoyting with warnings, don't worry. I will take over this one ;)
> > > 
> > > 
> > > Interesting, it was working for me. But I was using a newer version, maybe they changed something there.
> > 
> > 
> > Which version? The problem seems to be that is not detecting the system headers as headers thinking
> 
> It was the latest, which is probably some 2022 one. But it's very possible I missed something.

Turns out that is the MSVC version of the gh actions runners/dockers, whatever is in there. I've checked and it's version 17.x and I found this resource online: https://discourse.cmake.org/t/marking-headers-as-system-does-not-suppress-warnings-on-windows/6415 that says at least version 19

cpp/kiss_icp/core/Preprocessing.cpp
```diff
@@ -55,7 +55,8 @@ std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d>
     }
     std::vector<Eigen::Vector3d> frame_dowsampled;
     frame_dowsampled.reserve(grid.size());
-    for (const auto &[_, point] : grid) {
+    for (const auto &[voxel, point] : grid) {
+        (void)voxel;
```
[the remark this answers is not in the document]
nachovizzo: Sadly on older gcc this will fail with unused-variable' This is the case for ROS-melodic, and judging by how our community works, I know that if I break the build of ROS-melodic we will get issues automatically :) So I guess is worth the hack

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -138,6 +138,7 @@ std::vector<Eigen::Vector3d> VoxelHashMap::Pointcloud() const {
     std::vector<Eigen::Vector3d> points;
     points.reserve(max_points_per_voxel_ * map_.size());
     for (const auto &[voxel, voxel_block] : map_) {
+        (void)voxel;
         std::for_each(voxel_block.points.cbegin(), voxel_block.points.cend(),
```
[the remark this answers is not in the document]
nachovizzo: Indeed, you are right, I think it was leftovers from where I was trying to parallelize that region somehow (Usually I use std::for_each everywhere because of this)

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -138,6 +138,7 @@ std::vector<Eigen::Vector3d> VoxelHashMap::Pointcloud() const {
...
                       [&](const auto &point) { points.emplace_back(point); });
```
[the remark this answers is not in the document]
nachovizzo: I know it's a bad practice, but when I need to capture a few arguments, I get in lazy-mode. I know it's an excuse, though. Taking the push_back advice, thanks!

ahans:
> Turns out that is the MSVC version of the gh actions runners/dockers, whatever is in there. I've checked and it's version 17.x and I found this resource online: https://discourse.cmake.org/t/marking-headers-as-system-does-not-suppress-warnings-on-windows/6415 that says at least version 19

That link says you need to upgrade CMake. And some GH documentation says `windows-2022` has VS 2022. So this should be fine actually. Please don't merge this like this, we can do this proper. I can have a closer look tomorrow.

cpp/kiss_icp/cmake/CompilerOptions.cmake
```diff
@@ -22,9 +22,30 @@
...
+            # Clang/AppleClang
+            $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-fcolor-diagnostics>
+            $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Werror>
+            $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Wall>
+            $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Wextra>
+            $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Wconversion>
+            $<$<COMPILE_LANG_AND_ID:CXX,Clang,AppleClang>:-Wno-sign-conversion>
```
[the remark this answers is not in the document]
nachovizzo: Just copy-pasted from OpenVDB ;) I was trying to see what triggers. Honestly speaking, if people from OpenVDB took a different between clang and gcc flags, there might be a reasoning... I'd rather stick to it since I'm not that crazy about flags in general. But those added spitted some few more warnings here and there

nachovizzo:
> > Turns out that is the MSVC version of the gh actions runners/dockers, whatever is in there. I've checked and it's version 17.x and I found this resource online: https://discourse.cmake.org/t/marking-headers-as-system-does-not-suppress-warnings-on-windows/6415 that says at least version 19
> 
> That link says you need to upgrade CMake. And some GH documentation says `windows-2022` has VS 2022. So this should be fine actually. Please don't merge this like this, we can do this proper. I can have a closer look tomorrow.

On the last runs:
```sh
1s
Current runner version: '2.303.0'
Operating System
  Microsoft Windows Server 2022
  [1](https://github.com/PRBonn/kiss-icp/actions/runs/4536893530/jobs/7994133538#step:1:1)0.0.[2](https://github.com/PRBonn/kiss-icp/actions/runs/4536893530/jobs/7994133538#step:1:2)0[3](https://github.com/PRBonn/kiss-icp/actions/runs/4536893530/jobs/7994133538#step:1:3)[4](https://github.com/PRBonn/kiss-icp/actions/runs/4536893530/jobs/7994133538#step:1:4)8
  Datacenter
```

and
```
-- The CXX compiler identification is MSVC 19.34.31942.0
```

Nevertheless, it does fail anyway :S

ahans:
> Nevertheless, it does fail anyway :S

Yes, because you need to upgrade CMake to properly. That's at least what they say here:

> Ignoring warnings from system headers is only supported with MSVC with the Ninja generators as of CMake 3.22 and the **Visual Studio generators as of CMake 3.24**

Let me quickly try that.

nachovizzo:
> > Nevertheless, it does fail anyway :S
> 
> Yes, because you need to upgrade CMake to properly. That's at least what they say here:
> 
> > Ignoring warnings from system headers is only supported with MSVC with the Ninja generators as of CMake 3.22 and the **Visual Studio generators as of CMake 3.24**
> 
> Let me quickly try that.

Or use ninja, it does not fail in the windows-python-package because it is using ninja anywyas

ahans:
> > > Nevertheless, it does fail anyway :S
> > 
> > 
> > Yes, because you need to upgrade CMake to properly. That's at least what they say here:
> > > Ignoring warnings from system headers is only supported with MSVC with the Ninja generators as of CMake 3.22 and the **Visual Studio generators as of CMake 3.24**
> > 
> > 
> > Let me quickly try that.
> 
> Or use ninja, it does not fail in the windows-python-package because it is using ninja anywyas

The CMake upgrade did the trick. Anyways, keep whatever you want, we have some options now.

nachovizzo:
> > > > Nevertheless, it does fail anyway :S
> > > 
> > > 
> > > Yes, because you need to upgrade CMake to properly. That's at least what they say here:
> > > > Ignoring warnings from system headers is only supported with MSVC with the Ninja generators as of CMake 3.22 and the **Visual Studio generators as of CMake 3.24**
> > > 
> > > 
> > > Let me quickly try that.
> > 
> > 
> > Or use ninja, it does not fail in the windows-python-package because it is using ninja anywyas
> 
> The CMake upgrade did the trick. Anyways, keep whatever you want, we have some options now.

I tried some minutes the ninja thing and didn't work out great in my case... so dropping this for now. I will just change the cmake version to 3.24 since this will match the requirements of the project, and I don't want to change that now.

Thanks for the effort and the guidance ;)

cpp/kiss_icp/cmake/CompilerOptions.cmake
```diff
@@ -22,9 +22,29 @@
...
+            # GNU
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-fdiagnostics-color=always>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Werror>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wall>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wextra>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-pedantic>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wcast-align>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wcast-qual>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wconversion>
+            $<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wdisabled-optimization>
```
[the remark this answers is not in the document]
ahans: In that case I would prefer just taking the standard `-Wall -Wextra -Wpedantic`, but I'm also fine with this. Had a quick look at OpenVDB. It looks like they know what they're doing, so there are most probably no "bad" warning flags enabled here.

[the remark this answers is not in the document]
nachovizzo: oki doki. Then I will leave this here

nachovizzo:
- [ ] Fix ROS1 Melodic build

ahans:
> * [ ]  Fix ROS1 Melodic build

Not sure what you mean. It seems to be building fine for me. Maybe merge the other PR first and then we rebase this and I can see what CI has to say about it.

nachovizzo:
> > * [ ]   Fix ROS1 Melodic build
> 
> Not sure what you mean. It seems to be building fine for me. Maybe merge the other PR first and then we rebase this and I can see what CI has to say about it.

was a note for myself, actually. You can test the build in a dockerize environment using `gitlab-ci-local ros1_melodic` on the root repo (while we finish porting from gitlab CI to github CI)

Tool:https://github.com/firecow/gitlab-ci-local

PR #103, opened 2023-03-26
Move style checks to pre-commit

ahans:
This is roughly what I had in mind. It allows you to maintain only `.pre-commit-config.yaml` for all style checks. The GH action runs it in CI, manually you can just `pre-commit run [--all]` and don't have to manage tools or anything manually. It's all taken care of by pre-commit.

Just putting this up to show what I was talking about. No need to merge this if you don't like it.


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,23 @@
+name: Style Check
+
+on:
+  pull_request:
```
ahans, line 4: I change this so it runs on all PRs to be able to test this. I would revert this before merging.

.pre-commit-config.yaml
```diff
@@ -1,17 +1,21 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.4.0
```
ahans, line 3: Ran ` pre-commit autoupdate`  to update these. If for some reason you want a specific version (maybe clang-format needs to be version 14?), we can revert of course.

.pre-commit-config.yaml
```diff
@@ -1,17 +1,21 @@
...
+    rev: v16.0.0
     hooks:
     -   id: clang-format
 -   repo: https://github.com/psf/black
-    rev: 22.10.0
+    rev: 23.1.0
     hooks:
     -   id: black
         args: [--config=python/pyproject.toml]
+-   repo: https://github.com/ahans/cmake-format-precommit
```
ahans, line 18: Looked like a good chance to also include cmake-format, since you have a `.cmake-format.yaml`  in here (I didn't even know this exists, I'm not actively using CMake in any project). But it looks like neither the cmake-format project nor the pre-commit are very well maintained. cmake-format is missing some yaml dependency. The right solution would have been to fix that, but since then we'd also have to touch the hook, I forked the hook and put the dep there.

cpp/kiss_icp/3rdparty/sophus/sophus.cmake
```diff
@@ -31,11 +31,8 @@ ExternalProject_Add(
   CONFIGURE_COMMAND ""
   BUILD_COMMAND ""
   INSTALL_COMMAND ""
-  CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
-             -DSOPHUS_USE_BASIC_LOGGING=ON
-             -DBUILD_SOPHUS_EXAMPLES=OFF
-             -DBUILD_SOPHUS_TESTS=OFF
-             -DCMAKE_BUILD_TYPE=Release)
+  CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR> -DSOPHUS_USE_BASIC_LOGGING=ON -DBUILD_SOPHUS_EXAMPLES=OFF
```
ahans, line 34: This and other CMake related changes are due to cmake-format.

.github/workflows/pre-commit.yml
nachovizzo: 👌

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,23 @@
...
+  push:
+    branches: [main]
+
+jobs:
```
nachovizzo, line 8: I like the idea in general, is this a common practice nowadays? My only concern is that the main check is even more hidden inside layers of abstractions

But I agree that for each check we want to add is yet another action to configure (as an example back and clang format were different)

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: set PY
+        run: echo "PY=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
nachovizzo, line 22: I'm guessing this is to speed up the CI?

.pre-commit-config.yaml
nachovizzo: Not really, just copy pasted what was on the official documents

.pre-commit-config.yaml
```diff
@@ -1,17 +1,21 @@
...
-    rev: v3.2.0
+    rev: v4.4.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: check-added-large-files
 -   repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: 'v14.0.0'
+    rev: v16.0.0
```
nachovizzo, line 10: This I'd keep 14, just because is the default version of Ubuntu 22.04, it might break with newer versions

.pre-commit-config.yaml
nachovizzo: I also tried and was not working 😂 thanks for providing a solution!

cpp/kiss_icp/3rdparty/sophus/sophus.cmake
nachovizzo: Fine 👌

nachovizzo:
How does a failure case look like? Is it clear if the failure is on black or any formatter? Would you mind providing a screenshot or something similar

nachovizzo:
@ahans sorry I just merged main again into this branch. I keep accidentally pushing the damn button 😂. I still can't believe I can merge stuff into your private repo 😱. Sorry I did it twice 😔

ahans:
> How does a failure case look like? Is it clear if the failure is on black or any formatter? Would you mind providing a screenshot or something similar

Here's an example where I messed up the formatting of some Python file:

image

It checks and fixes in one go, so next time I run it everything would be green again. That's something I disliked at first. I would have preferred some "check only" mode. But having used it for a while it's actually not an issue at all in my workflow. For most file types I have some "format on change" enabled anyways, but occasionally it does happen that something does not get formatted properly. Then I see that the "pre-commit" stage in CI is red and I want to run "fix" locally anyways. Really can't think of a reason why I would want a "check" only. It would always be followed by a "fix" anyways.

ahans:
> @ahans sorry I just merged main again into this branch. I keep accidentally pushing the damn button joy. I still can't believe I can merge stuff into your private repo scream. Sorry I did it twice pensive

No worries, I left the option checked that allows you (or any other maintainer) to make edits. I think it's fairly reasonable. Otherwise you would have to rely on me for everything (or copy my branch into your own and open up another PR).

nachovizzo:
> > @ahans sorry I just merged main again into this branch. I keep accidentally pushing the damn button joy. I still can't believe I can merge stuff into your private repo scream. Sorry I did it twice pensive
> 
> No worries, I left the option checked that allows you (or any other maintainer) to make edits. I think it's fairly reasonable. Otherwise you would have to rely on me for everything (or copy my branch into your own and open up another PR).

Deswegen! Now I get it. Much easier to work like this. Thanks!

.github/workflows/pre-commit.yml
ahans: It's true that it takes some getting used to. I first saw it used in some repo here on GH. It seemed interesting, so I started looking into it. And now it's used in all repos I maintain. No complaints so far, and actually at least one praise where a colleague said it's great that CI tells them exactly what to do. This is an example of a failure report in CI:

image

The syntax of the `.pre-commit-config.yaml` is a bit obscure, same with the GH actions (or `.gitlab-ci.yml` for that matter), but it's nothing that everyone must understand and know how to modify.

.github/workflows/pre-commit.yml
ahans: Yes, it just caches the stuff that pre-commit downloads. I think it creates a venv for each hook. The key makes sure that in case the Python version changes or `.pre-commit-config.yaml`, we don't use the cache, but regenerate. This helps in CI to avoid clashes between different branches that change any of the two. If they all hit the same cache entry, they would keep invalidating the cache.

.pre-commit-config.yaml
ahans: Alright, will revert it then.

.github/workflows/pre-commit.yml
nachovizzo: Ok awesome, when I was wondering about the failure example was talking about the CI. So, they even tell you how to run it locally, this is the best of the world in my opinion. Usually people will come back with the error msg!

When I tried pre-commit 4 years ago it was not that developed, and thus I dropped it, it all seemed too chunky to maintain. But just putting it here to simplify the CI(that among all the examples you gave, the syntax that suck the most is the GH actions) it's more than worth it! I love the idea ;) Specially because pre-commit must NOT be enforced for developers (which was my prev intuition)

nachovizzo:
> > How does a failure case look like? Is it clear if the failure is on black or any formatter? Would you mind providing a screenshot or something similar
> 
> Here's an example where I messed up the formatting of some Python file:
> 
> image
> 
> It checks and fixes in one go, so next time I run it everything would be green again. That's something I disliked at first. I would have preferred some "check only" mode. But having used it for a while it's actually not an issue at all in my workflow. For most file types I have some "format on change" enabled anyways, but occasionally it does happen that something does not get formatted properly. Then I see that the "pre-commit" stage in CI is red and I want to run "fix" locally anyways. Really can't think of a reason why I would want a "check" only. It would always be followed by a "fix" anyways.

If you do `git commit` locally, it will stop without changing the code. Assuming you've installed the hook `pre-commit install` Which I find terrific.

ahans:
> If you do `git commit` locally, it will stop without changing the code. Assuming you've installed the hook `pre-commit install` Which I find terrific.

I've never run `pre-commit install`. Been always a bit skeptical about git hooks. Maybe I should try it out! ;-)

> I think is ready to merge @ahans . If you agree let me know because I need to change some config on the GH side

I reverted the clang-format version now and also changed the config for when to run the workflow to match the other ones. So yes, from my end I would say it's ready.

ahans:
I also rebased onto latest `main` and dropped your merge commit. Just make sure to squash the two fixups into the one commit. But I think that happens automatically anyway.

nachovizzo:
> I also rebased onto latest `main` and dropped your merge commit. Just make sure to squash the two fixups into the one commit. But I think that happens automatically anyway.

yup, cuz I squash and merge always anyways

.github/workflows/pre-commit.yml
nachovizzo: Btw, the "If you are seeing this message in CI" is gone in newer versions apparently :cry:

PR #105, opened 2023-03-27
Build ROS 2 node in CI

ahans:
Here's a GH action that builds the ROS 2 node. If you want to support other distributions also, could make sense to introduce an extra `ros.yml` for this.


.github/workflows/cpp.yml
```diff
@@ -25,3 +25,12 @@ jobs:
         run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} ${{github.workspace}}/cpp/kiss_icp
       - name: Build
         run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}}
+
+  ros2_node:
```
nachovizzo, line 29: Finally, an example that looks familiar to my eyes (docker-based). Thanks!

nachovizzo:
@ahans Maybe adding the other 3 distros is not a big deal?

```yaml
ros1_melodic:
  image: osrf/ros:melodic-desktop-full
  stage: build
  before_script:
    - apt-get update && sudo apt-get install -y gpg wget
    - wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
    - echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ bionic main' | sudo tee /etc/apt/sources.list.d/kitware.list >/dev/null && apt update
    - rm /usr/share/keyrings/kitware-archive-keyring.gpg
    - apt-get install -y kitware-archive-keyring
    - apt-get install -y cmake
  script:
    - rm -rf ~/catkin_ws/
    - mkdir -p ~/catkin_ws/
    - cp -r ${CI_PROJECT_DIR} ~/catkin_ws/src
    - cd ~/catkin_ws/
    - catkin_make

ros1_noetic:
  image: osrf/ros:noetic-desktop-full
  stage: build
  before_script:
    - rm -rf ~/catkin_ws/
    - mkdir -p ~/catkin_ws/
    - cp -r ${CI_PROJECT_DIR} ~/catkin_ws/src
  script:
    - cd ~/catkin_ws/
    - catkin_make

ros2_foxy:
  image: osrf/ros:foxy-desktop
  stage: build
  script:
    - colcon build --event-handlers console_direct+
```

nachovizzo:
> > @ahans Maybe adding the other 3 distros is not a big deal?
> 
> I don't care about the others and wanted to leave some work for you. 😀 Anyways, learned some more about GH actions and matrix strategy. I think it's done now and turned out quite ok.

Agree, this turned out to be nice and simple.

Any idea why people would opt for [This](https://github.com/ros-tooling/action-ros-ci) instead of the plain docker syntax you used? I found yours much less confusing, but maybe I'm way too used to Docker 🤷

Anyway, this looks good to me. Meanwhile I'm trying to analyze why the CI is not failing for ROS1 Melodic 🤔...

ahans:
> Any idea why people would opt for [This](https://github.com/ros-tooling/action-ros-ci) instead of the plain docker syntax you used? I found yours much less confusing, but maybe I'm way too used to Docker shrug

It looks like this does a bit more than the equivalent of pulling and running a Docker image. It can checkout other dependent repos and run `rosdep` automatically, supposedly reading the deps from `package.xml` files. In the Docker-based approach we would have to do that manually. For the present project I would still prefer the Docker approach, because there's less magic happening and you can easily repro locally exactly what the workflow does. You can probably also run an action locally, but that's another tool you need to install and understand how to use ...

.github/workflows/ros.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        release: [melodic, noetic]
+    container: osrf/ros:${{ matrix.release }}-desktop-full
+    steps:
+      - name: Setup cmake
+        uses: jwlawson/actions-setup-cmake@v1.13
+        with:
+          cmake-version: "3.24.x"
```
nachovizzo: This action was failing on ROS1 because `node` was missing. I also tried to install node with another action but it was also failing, so I basically copy pasted what I had in the gitlab-ci in here to make it work. If you want to try a better solution go ahead, if not let me know so I can merge the latest change ;)

nachovizzo: Strange, is not working when using `act` but it does work on the real CI. Sorry for the noise

PR #107, opened 2023-03-28
More robust detection of ROS 1 vs ROS 2

jlblancoc:
With this change, one can build ROS 1 using colcon or catkin_make.

For the records: Using "ROS_VERSION" here requires adding the dependency ros_environment to package.xml for that env variable to exist on build farms too.


nachovizzo:
Thanks for the contribution @jlblancoc ! I will merge first the PR that add ROS to the CI (#105) and then merge this right away

PR #109, opened 2023-03-30
Nacho/improve mcap dataloader

nachovizzo:
Using the feedback from https://github.com/orgs/foxglove/discussions/671 I've improved the mcap dataloader to be completely independent of ROS2. The topic-guess utility still works.

This PR continues the work started at #104


python/kiss_icp/datasets/mcap.py
```diff
@@ -104,29 +61,37 @@ def __getitem__(self, idx):
     def __len__(self):
         return self.n_scans
 
+    def _get_n_scans(self) -> int:
+        return [
+            count
+            for (id, count) in self.summary.statistics.channel_message_counts.items()
+            if self.summary.channels[id].topic == self.topic
+        ][0]
```
jtbandes, line 69: Might want to use `sum(...)` instead of `[...][0]`?

nachovizzo: This is a tricky one... Is possible to have multiple channels with the same topic? the `sum` keyword give me the wrong impression, that we are accumulating the number of messages.

The current solution looks horrible anyway, this I agree :rofl:

jtbandes: Yes, it's allowed, maybe more likely with conversion from ROS 1 bags where each connection is converted to a channel. I'm not sure off the top of my head if ROS 2 recordings consolidate topics.

nachovizzo: Then the sum thing make sense, would you mind opening a PR, so I can give "some credit"? Otherwise, I can also do it, just let me know

jtbandes: https://github.com/PRBonn/kiss-icp/pull/114

nachovizzo: Thanks!

PR #114, opened 2023-03-31
Use sum() over channel_message_counts

jtbandes:
See https://github.com/PRBonn/kiss-icp/pull/109#pullrequestreview-1366020132


nachovizzo:
Thanks!

PR #120, opened 2023-04-03
Add timestamps - KITTI, Rosbag

### LINKED ISSUE #112: Add scan-timestamp for dataloaders we know already this value (TUM poses format)
We added the TUM pose format to the results of the python pipeline (#45) But we haven't extended this to all other existing datasets besides tum.

The idea is rather simple, take as an example the rosbag, rosbag2, mcap dataloaders. For each scan, we already know the timestamp of that particular sensor frame (not to be confused with point-timestamps). This is the time we recorded the scan.

We should store that value for each scan, and take advantage of the tum poses format to produce a result which is more meaningful for some applications.

saurabh1002:
Partially fix #112 

Add timestamps for:

- KITTI Odometry dataloader
- ROS1 rosbag dataloader


python/kiss_icp/datasets/kitti.py
```diff
@@ -101,4 +105,4 @@ def read_calib_file(file_path: str) -> dict:
                     # The format in KITTI's file is <key>: <f1> <f2> <f3> ...\n -> Remove the ':'
                     key = tokens[0][:-1]
                     calib_dict[key] = values
-        return calib_dict
```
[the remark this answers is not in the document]
saurabh1002: Done

python/kiss_icp/datasets/kitti.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, data_dir, sequence: int, *_, **__):
 
         self.scan_files = sorted(glob.glob(self.velodyne_dir + "*.bin"))
         self.calibration = self.read_calib_file(os.path.join(self.kitti_sequence_dir, "calib.txt"))
+        self.timestamps = self.read_timestamps(os.path.join(self.kitti_sequence_dir, "times.txt"))
```
nachovizzo, line 37: Why we need `read_timestamps` at all? I'd just put the `np.loadtxt` command here and remove the other function (unless the pipeline API asks for it)

python/kiss_icp/datasets/rosbag.py
```diff
@@ -55,8 +56,12 @@ def __len__(self):
 
     def __getitem__(self, idx):
         # TODO: implemnt [idx], expose field_names
```
saurabh1002, line 58: The read_messages function in rosbag library returns a Generator Object. It is kind of impossible to implement an indexing based access unless we convert it to a list implicitly.

Exposing the field names is trivial though, we just need to store/return the `msgs.fields` output which is a list of strings containing details of each field in the PointCloud2 message.

nachovizzo: I see, I actually forgot why I wanted to expose the field_names.

About the sequential thing, I understand. Is not possible, and is not a dream of mine anymore :)

PR #121, opened 2023-04-04
Remove clang hacks

### LINKED ISSUE #118: Attempt to fix clang compiler hack
Introduced the following hacks, just to compile with clang compilers (basically macOS):

https://github.com/PRBonn/kiss-icp/blob/52a99d034318241d354ab2a50c7924572de70314/cpp/kiss_icp/core/VoxelHashMap.cpp#L103

https://github.com/PRBonn/kiss-icp/blob/52a99d034318241d354ab2a50c7924572de70314/cpp/kiss_icp/core/VoxelHashMap.cpp#L121

The idea of this issue is to remove that hack and beautify the code.

Some resources online:
- https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding
- https://youtrack.jetbrains.com/issue/CPP-29193/Erroneous-error-highlighting-reference-to-local-binding-variable-declared-in-enclosing-function
- https://github.com/doctest/doctest/issues/279

ahans:
The first instance of the clang workaround we can remove by simply dropping the need for a lambda: `std::for_each` with a capture-all lambda doesn't add anyting over a range-based for loop. In the second instance the workaround wasn't even necessary.

Closes #118

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -96,38 +96,29 @@ VoxelHashMap::Vector3dVectorTuple VoxelHashMap::GetCorrespondences(
         tbb::blocked_range<points_iterator>{points.cbegin(), points.cend()},
         // Identity
         ResultTuple(points.size()),
-        // 1st lambda: Paralell computation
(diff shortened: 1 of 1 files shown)


nachovizzo:
Terrific solution!

Thanks for fixing this horrible mistake :D . I'm testing locally and merge afterward

Issue #122, opened 2023-04-04
poses vector grows indefinitely


ahans:
The [poses vector](https://github.com/PRBonn/kiss-icp/blob/main/cpp/kiss_icp/pipeline/KissICP.hpp#L78) is only ever added to. While poses are small objects, eventually any system will run out of memory. We need some bound for this.

To fix this, we could define some number of last poses that are kept. However, it looks like currently only the very first and last two poses are ever used. An even better fix might therefore be just keeping those.

tizianoGuadagnino:
Hi @ahans,
The unbounded growth of the pose vector it is not a sensible problem in our point of view. A pose using Sophus is represented by 7 doubles. Unless there are specific memory restrictions, you need an extremely high number of poses for this to be a real problem. Furthermore, we need to store the all trajectory so that it can be inspected after a run.

ahans:
Keeping all poses is just nothing we need in KISS-ICP core. If you want to inspect them after a run, there should be another place where you accumulate the poses. Maybe somewhere in the Python driver. I was under the impression you wanted to do things properly in this project, and unbounded growth is simply tech debt. And in the present case it's actually quite simple to fix it, as shown in my PR #124. Granted, the Python side of things would need some work, but it's still small enough to not accept this. But all of that is IMHO. If you want to keep this as is, fine.

PR #124, opened 2023-04-04
Only keep first and last two poses

ahans:
This is how the suggestion from #122 could be implemented. Maybe you can think of something even easier. For example, I'm not sure if we really have to keep the first pose in order to determine if we have moved. I tried to retain the existing behavior for now.


ahans:
I already know that this breaks the Python code. Without tests and no need to build Python, I didn't notice this. I also found it a bit surprising that so much functionality is implemented in both, Python and C++. I would have expected that for Python we have a very thin wrapper. Anyway, I can of course also adapt the Python code, but would want to know first if the approach does even make sense. Thanks!

PR #126, opened 2023-04-05
Use official TBB v2021.8.0 release

### LINKED ISSUE #113: Upgrade tbb to an official version and drop nacho's fork
The main problem was that some time ago, tbb was not supporting (or I could not find) a CMake-based build system. That's why I built my own [fork](https://github.com/nachovizzo/tbb) on top of someone else fork.

Of course, this is not ideal, and it would be great to upgrade to the latest stable release from [oneAPI/tbb](https://github.com/oneapi-src/oneTBB). In the main CMakelist, they recommend NOT using tbb as static library. It would be interesting to know why. If we discover that indeed this is a concern, then we should consider adding the shared object in the python wheels, as many other libraries do (Open3D, pytorch, etc). This is one of the many reasons of why it is NOT a good idea to maintain your own fork of a library. :)

Be aware, that now the library is part of the OneAPI ecosystem and the API changed substantially.

ahans:
The latest stable release seems to be working just fine?

I'm not sure if this fixes #113. The issue with using discouraged static linking is still open. But maybe worth creating an extra ticket for this?


nachovizzo:
Strange… last time I've checked, cmake was not supported. But looks this is now supported. Terrific news!

I agree that the static linking should be investigated separately given that there was such an easy fix for the upstream. I will debug the produced symbols for a bit and merge this afterward.

nachovizzo:
@ahans When I run  `$ cibuildwheel --platform linux --output-dir wheelhouse ./python/`:
 locally, I get: `  /opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: cannot find -ltb`

Have you seen this problem? On the CI looks to be working :thinking: , but this is now suspicious

ahans:
> @ahans When I run `$ cibuildwheel --platform linux --output-dir wheelhouse ./python/`: locally, I get: ` /opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: cannot find -ltb`
> 
> Have you seen this problem? On the CI looks to be working thinking , but this is now suspicious

I don't know `cibuildwheel`, but can have a look. But exactly does "running locally" mean in your case?

nachovizzo:
> > @ahans When I run `$ cibuildwheel --platform linux --output-dir wheelhouse ./python/`: locally, I get: ` /opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: cannot find -ltb`
> > Have you seen this problem? On the CI looks to be working thinking , but this is now suspicious
> 
> I don't know `cibuildwheel`, but can have a look. But exactly does "running locally" mean in your case?

So `cibuildwheel` is the tool I use to, well, build the python wheels. It's like an automation tool, that use docer of course, to build Python wheel that requiere some binary code.

to reproduce, just `pip install cibuildwheel` and then run the command above. This will launch the build process, for the different python versions specified in the pyproject.tom file

ahans:
> to reproduce, just `pip install cibuildwheel` and then run the command above. This will launch the build process, for the different python versions specified in the pyproject.tom file

Alright, I was able to repro. Looks like on that RedHat-based distro libs go into `lib64` instead of `lib`. Not sure how CMake figures that out. We could probably use the same logic in `tbb.cmake`. I now just added `lib64` to the lib directories. I'd say that's fine.

ahans:
Looks like I found a nicer way now. I guess `CMAKE_INSTALL_LIBDIR` is also what TBB uses and your custom TBB was not 100% correct set up in that respect.

ahans:
Oh well, now the Ubuntu builds fail. Going back to listing both, `lib` as well as `lib64`.

cpp/kiss_icp/3rdparty/tbb/tbb.cmake
```diff
@@ -43,6 +43,6 @@ set(TBB_ROOT ${INSTALL_DIR} CACHE INTERNAL "TBB_ROOT Install directory")
 add_library(TBBHelper INTERFACE)
 add_dependencies(TBBHelper external_tbb)
 target_include_directories(TBBHelper SYSTEM INTERFACE ${INSTALL_DIR}/include)
-target_link_directories(TBBHelper INTERFACE ${CMAKE_INSTALL_LIBDIR})
+target_link_directories(TBBHelper INTERFACE ${INSTALL_DIR}/lib ${INSTALL_DIR}/lib64)
```
nachovizzo, line 46: This was actually the propper way of doing it. Some stuff was missing though

1. `include(GNUInstallDirs)`
1. `${INSTALL_DIR}/${CMAKE_INSTALL_LIBDIR}`

`CMAKE_INSTALL_LIBDIR` only spits `lib` or `lib64`. We need to preppend our custom, local direcotry, `INSTALL_DIR`

nachovizzo:
@ahans Merging this so I can create a new release to debug the release pipeline after this change. Thanks again!

Issue #127, opened 2023-04-05
Dynamically link tbb library


nachovizzo:
This continues the saga of #113 :

The main idea is to drop the `-DBUILD_SHARED_LIBS=OFF` prefix in the tbb external cmake script. The problem would be is that we would need to fetch that shared library during build time `libtbb.so` and add it to the python wheels. This is a common practice in many libraries, so I guess it shouldn't be a big problem

nachovizzo:
Great!

I good starting point is to check my [lecture on build systems]( https://www.youtube.com/watch?v=seW8wn4-EfI&list=PLgnQpQtFTOGRv7VS6fYerEbT4ckBovKur&index=2)  if your already have this knowledge let me know so I can provide the next pointers

nachovizzo:
Won't fix

PR #142, opened 2023-04-12
Nacho/tbb as shared library

nachovizzo:
First approximation to solve #127


nachovizzo:
This actually works, and also inject the .so in the wheel with the right rpath. But for some reason., cibuildwheel is not happy about this fact
```
  ValueError: Cannot repair wheel, because required library "libtbb.so.12" could not be located

§§§

PR #149, opened 2023-04-26
Add options to toggle odom and alias tf (continued from #92)

willcbaker:
See [#92](https://github.com/PRBonn/kiss-icp/pull/92)

This feature introduces parameters to toggle on/off publishing of TFs to avoid tf conflicts from other sources.

For examples:

if this were used as an odometry signal in http://wiki.ros.org/robot_localization you need to toggle off the odom->base_link tf
if not using base link frame_id but the tf already exists (e.g. from robot_state_publisher) do not publish an identity transform


willcbaker (commented):
clean whitespace

nachovizzo:
@willcbaker would you mind giving me access to your branch so I can directly make the changes there?

PR #158, opened 2023-05-13
Added support to save poses in ros1 for evaluation in tum and kitti format

sumanthrao1997:


sumanthrao1997:
I plan to do in ros2 as well soon, I will let you know once its done

nachovizzo:
@sumanthrao1997 please re-open when the ROS 2 service is also ready. Thanks :)

Issue #162, opened 2023-05-18
Make ROS2 OdometryServer composable


roncapat:
It shall be easy to transform OdometryServer in a ROS2 component. Is this way, you could provide:

- a standalone node (as it is now)
- a component

... at the same time.

Moreover, the component would benefit of much higher transfer rate of PointClouds messages thanks to intra-process communication.

Would you like a patch to provide this?

nachovizzo:
@roncapat Thanks for the suggestion. As I read this, I get to know about components in ROS 2. So, I'm literally blind to this topic. I'm more than willing to help review a merge request about this.

Would you be interested in doing so? If not, could you provide some pointers on where to start and what example to look in order to achieve this?

Thanks!

roncapat:
If you'd like to know more about composition, a paper has been released this week :)
https://arxiv.org/abs/2305.09933

nachovizzo:
Just saw the post yesterday on LinkedIn. Amazing works from Steve as usual

PR #163, opened 2023-05-19
As ROS2 rclcpp Component [#162]

roncapat:
See [#162].

This produced a new target: `odometry_component`, while `odometry_node` is then automagically generated by `rclcpp_components_register_node` CMake macro.

### DIFF
--- a/ros/CMakeLists.txt
+++ b/ros/CMakeLists.txt
@@ -75,16 +75,22 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
   add_library(odometry_component SHARED ros2/OdometryServer.cpp)
   target_include_directories(odometry_component PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include)
   target_link_libraries(odometry_component kiss_icp::pipeline)
-  ament_target_dependencies(odometry_component rcutils rclcpp rclcpp_components nav_msgs sensor_msgs tf2_ros)
+  ament_target_dependencies(
+    odometry_component 
+    rcutils 
+    rclcpp 
+    rclcpp_components 
+    nav_msgs 
+    sensor_msgs 
+    tf2_ros)
 
-  rclcpp_components_register_node(
-    odometry_component PLUGIN "kiss_icp_ros::OdometryServer" EXECUTABLE odometry_node)
+  rclcpp_components_register_node(odometry_component PLUGIN "kiss_icp_ros::OdometryServer" EXECUTABLE odometry_node)
 
   install(TARGETS odometry_component LIBRARY DESTINATION lib RUNTIME DESTINATION lib/${PROJECT_NAME})
   install(DIRECTORY launch rviz DESTINATION share/${PROJECT_NAME}/)
 
   ament_package()
 
-  else()
+else()
   message(FATAL_ERROR "catkin or colcon not found KISS-ICP-ROS disabled")
 endif()
--- a/ros/CMakeLists.txt
+++ b/ros/CMakeLists.txt
@@ -65,18 +65,26 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
 
   find_package(ament_cmake REQUIRED)
   find_package(nav_msgs REQUIRED)
+  find_package(rcutils REQUIRED)
   find_package(rclcpp REQUIRED)
+  find_package(rclcpp_components REQUIRED)
   find_package(sensor_msgs REQUIRED)
   find_package(tf2_ros REQUIRED)
 
   # ROS2 node
-  add_executable(odometry_node ros2/OdometryServer.cpp)
-  target_include_directories(odometry_node PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include)
-  target_link_libraries(odometry_node kiss_icp::pipeline)
-  install(TARGETS odometry_node RUNTIME DESTINATION lib/${PROJECT_NAME})
+  add_library(odometry_component SHARED ros2/OdometryServer.cpp)
+  target_include_directories(odometry_component PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include)
+  target_link_libraries(odometry_component kiss_icp::pipeline)
+  ament_target_dependencies(odometry_component rcutils rclcpp rclcpp_components nav_msgs sensor_msgs tf2_ros)
+
+  rclcpp_components_register_node(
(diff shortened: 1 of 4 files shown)


ros/package.xml
```diff
@@ -44,6 +44,9 @@
   <exec_depend condition="$ROS_VERSION == 2">ros2launch</exec_depend>
 
   <!-- ROS1/2 dependencies -->
+  <depend>rcutils</depend>
```
nachovizzo, line 47: I guess this line can be moved to the "ROS 2 only" dependencies:
```xml
  <!-- ROS2 dependencies -->
  <buildtool_depend condition="$ROS_VERSION == 2">ament_cmake</buildtool_depend>
  <depend condition="$ROS_VERSION == 2">rclcpp</depend>
  <depend condition="$ROS_VERSION == 2">rclcpp_components</depend>
  <exec_depend condition="$ROS_VERSION == 2">ros2launch</exec_depend>
```

The other two rclcpp,  rclcpp_components are already there

ros/CMakeLists.txt
```diff
@@ -65,18 +65,32 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
...
   find_package(ament_cmake REQUIRED)
   find_package(nav_msgs REQUIRED)
+  find_package(rcutils REQUIRED)
   find_package(rclcpp REQUIRED)
+  find_package(rclcpp_components REQUIRED)
   find_package(sensor_msgs REQUIRED)
   find_package(tf2_ros REQUIRED)
 
   # ROS2 node
-  add_executable(odometry_node ros2/OdometryServer.cpp)
```
nachovizzo, line 73: I'm new to this. This "component" can also be launched as a "normal node"? Is this backward compatible with the existing launch files?

ros/ros2/OdometryServer.hpp
```diff
@@ -73,3 +73,11 @@ class OdometryServer : public rclcpp::Node {
 };
 
 }  // namespace kiss_icp_ros
+
+#include "rclcpp_components/register_node_macro.hpp"
+
+// Register the component with class_loader.
+// This acts as a sort of entry point, allowing the component to be
+// discoverable when its library is being loaded into a running process.
+// is being loaded into a running process.
```
nachovizzo, line 82: "is being loaded into a running process." is duplicated ;)

ros/ros2/OdometryServer.hpp
```diff
@@ -37,7 +37,7 @@ namespace kiss_icp_ros {
 class OdometryServer : public rclcpp::Node {
 public:
     /// OdometryServer constructor
-    OdometryServer();
```
nachovizzo, line 40: Should this be deleted? Otherwise, the compiler will generate the default constructor. Or does this even matter for the component's API?

ros/ros2/OdometryServer.cpp
```diff
@@ -171,10 +172,3 @@ void OdometryServer::RegisterFrame(const sensor_msgs::msg::PointCloud2::SharedPt
     map_publisher_->publish(utils::EigenToPointCloud2(odometry_.LocalMap(), local_map_header));
 }
 }  // namespace kiss_icp_ros
-
-int main(int argc, char **argv) {
```
nachovizzo, line 175: Seeing this go away makes me very happy

ros/package.xml
roncapat: Ah, definetly not noticed them! I also saw the other comments, will reply / fix in the afternoon (in the meantime, one of my colleagues is testing this branch in a composed setup).

ros/CMakeLists.txt
roncapat: I'll leave some details here :)

Basically, your class becomes a component, which can be dynamically loaded by a ComponentManager. So that's why I renamed the target and changed from `add_executable` to `add_library`, and also removed the `main()`. Then, `rclcpp_components_register_node` macro in CMake generates a second target (using a default `main` routine) called like the old node. This target is, like the node you were building prior to this patch, a standalone executable that behaves exactly like a node. The only thing is, now it is dynamically linked to the component target (no user will notice this).

roncapat: So yes, it shall be compatible with existing launchfiles (if you can, have a try).

ros/ros2/OdometryServer.hpp
roncapat: For components API, the only requirement is to have the constructor taking node options. You mean explicitly `= delete;` the default one?

roncapat: Ah, another requirements to leverage true zero-copy is to use ConstSharedPtr instead of SharedPtr of ROS2 messages in the callback signature. I'll push the patch now.

roncapat: Also, I checked but the default constructor is automatically "deleted" since there is an user-defined constructor (https://en.cppreference.com/w/cpp/language/default_constructor)

roncapat:
@nachovizzo another thing... to reduce overhead and really leverage IPC while publishing PointClouds, the code shall instantiate PointClouds as `std::unique_ptr`s and then moved to the IPC manager by doing `publish(std::move(your-point-cloud-here))`. This may required some additional work around `util.hpp` file, so we may want to tackle it in a separate issue/PR later on IMHO.

nachovizzo:
> @nachovizzo another thing... to reduce overhead and really leverage IPC while publishing PointClouds, the code shall instantiate PointClouds as `std::unique_ptr`s and then moved to the IPC manager by doing `publish(std::move(your-point-cloud-here))`. This may required some additional work around `util.hpp` file, so we may want to tackle it in a separate issue/PR later on IMHO.

I guess it makes sense to do it on a separate MR. Great suggestion

nachovizzo:
> Let me know if the open points were all addressed from me or not. We're also doing some testing before recommending a merge.
> 
> EDIT: tested successfully on our side

Everything looks good to me. I won't be able to test this myself until next Thursday. So I guess it will be blocked, at least, until then.

Thanks again for putting the effort into this! highly appreciatged

roncapat:
Nachovizzo did you have the chance to test?

nachovizzo:
@roncapat not really! Sorry for the slowness, this week we are also at ICRA. So likely to test it next week. But I'm definitely going to look at it!

nachovizzo:
@roncapat I Tested in 3 datasets on my end, and works like a charm!

Have you encountered any issue? If not, I Will merge it!!

nachovizzo (approved):

PR #169, opened 2023-05-25
Account for multiple timestamp datatypes for PointCloud2 messages

nachovizzo:
Create a std::variant to account for different possible types in the timestamp channel for the PointCloud2 msg.

This PR add c++20 for the ROS wrappers


tizianoGuadagnino:
> Create a std::variant to account for different possible types in the timestamp channel for the PointCloud2 msg.
> 
> This PR add c++20 for the ROS wrappers

Remove the std::variant and use a templated lambdas for readability

tizianoGuadagnino (approved):

Issue #164, opened 2023-05-19
Reduce overhead and really leverage IPC while publishing PointClouds


nachovizzo:
@nachovizzo another thing... to reduce overhead and really leverage IPC while publishing PointClouds, the code shall instantiate PointClouds as `std::unique_ptr`s and then moved to the IPC manager by doing `publish(std::move(your-point-cloud-here))`. This may required some additional work around `util.hpp` file, so we may want to tackle it in a separate issue/PR later on IMHO.

_Originally posted by @roncapat in https://github.com/PRBonn/kiss-icp/issues/163#issuecomment-1554513658_

nachovizzo:
Closed in #171

PR #171, opened 2023-06-06
Enable zero-copy

roncapat:
See #164.

For now, only clouds and odometry ownership is moved to IPC manager.
Path publishing for now is kept standard (of course we may create a copy of it and gift to IPC manager).

UNTESTED CODE (this is why it is still draft)


roncapat:
@nachovizzo this seems working for us (cc. @fixit-davide). Could you run again your demo bags as for the previous PR? Theoretically, no need to compose the node to demonstrate that the patch is working, just run it standalone or as you like. 

Of course, you should see performance improvements in the following conditions:
- composed kiss-icp with two subscribers on the same topic (e.g. outgoing point cloud)
- intra-process enabled

roncapat:
Well... there is a little problem with intra-process enabled. TF2 StaticTransformBroadcaster does not support intra-process on. See my previous patch against TF2 [here](https://github.com/ros2/geometry2/pull/572). I will open a PR against them, but in the meantime I will also need to put in place a workaround in this KISS-ICP patch to remove the problem locally.

roncapat:
@nachovizzo have you found time to review? BTW, do you think we still need to support Foxy?

nachovizzo:
@roncapat not really! But it's on my todo-list. I don't think I will have even 5 minutes to check this on the next 2 weeks or so. Apologize for the slowness :(


But definitely going to look for it.


PS: What would be the advantage of killing foxy support?

roncapat:
No worries, just checking if I can do any improvements in the meantime :)

Foxy does not have QoS configurability (override support) so as you see the CI job is failing. 

QoS override is unfortunately needed to force-disable intra-process on the StaticTransformPublisher. If we don't suppress it, we can't use true intra-process all around the node. I'm trying to figure out if a fix for foxy could exist...

roncapat:
Hey @nachovizzo, do you have any updates? :)

nachovizzo:
@roncapat I guess I will kill ROS2 foxy, since EOL was reached a few weeks ago anyways. I'm reviewing this, testing it, and merging it soon.

Could you give me access to push to your branch? This way you don't need to change everything I'm asking

roncapat:
Of course, I sent you the invite :)

ros/ros2/Utils.hpp
```diff
@@ -154,19 +158,20 @@ std::vector<Eigen::Vector3d> PointCloud2ToEigen(const PointCloud2 &msg) {
...
 
-PointCloud2 EigenToPointCloud2(const std::vector<Eigen::Vector3d> &points, const Header &header) {
-    PointCloud2 msg = CreatePointCloud2Msg(points.size(), header);
-    FillPointCloud2XYZ(points, msg);
-    return msg;
+PointCloud2::UniquePtr EigenToPointCloud2(const std::vector<Eigen::Vector3d> &points,
+                                          const Header &header) {
+    auto msg = CreatePointCloud2Msg(points.size(), header);
+    FillPointCloud2XYZ(points, *msg);
+    return std::move(msg);
```
nachovizzo, line 165: Do we actually need to cast to rvalue here? You are using it anyways when you call this utility function.

nachovizzo (commented):
This looks good to me. I only have 1 small question.

ros/ros2/Utils.hpp
roncapat: Yes, it may be overkill (useless). In order to be sure, I've found this quick ref:
https://stackoverflow.com/questions/4316727/returning-unique-ptr-from-functions

nachovizzo: Ok, then I'm removing it (among with small changes).

I'll let you know when I finish so you give it a quick look ;)

roncapat: Nice :)

nachovizzo:
@roncapat I'm done reviewing and testing this. Thanks a lot again!

If you have time, just eye-ball the last diff to make sure we are on the same page, I'll merge after you give me thumbs up :+1: 

Thanks!

roncapat:
LGTM :)

nachovizzo:
Great!

Sorry I squeezed some small non-related changes :see_no_evil:

PR #177, opened 2023-06-22
Update to latest tsl_robin version

ybrunken:
There was a bug in `tsl_robin` leading to an unbounded memory consumption in rare situations. See: https://github.com/Tessil/robin-map/issues/52
This was fixed in version 1.2.0: https://github.com/Tessil/robin-map/releases/tag/v1.2.0


nachovizzo:
Thanks! I'll give it a quick test just in case and merge it :)

PR #178, opened 2023-06-26
Update hash function to use primes

ybrunken:
19349663 is not a prime. `19349663 = 41×471943`

The number set is like the one proposed in "[Optimized Spatial Hashing for Collision Detection of Deformable Objects](https://www.researchgate.net/publication/2909661_Optimized_Spatial_Hashing_for_Collision_Detection_of_Deformable_Objects)". Also here `19349663` is stated to be prime.
In "[Real-time 3D Reconstruction at Scale using Voxel Hashing](https://niessnerlab.org/papers/2013/4hashing/niessner2013hashing.pdf)", the non-prime is exchanged with `19349669`

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.hpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.hpp
@@ -48,7 +48,7 @@ struct VoxelHashMap {
     struct VoxelHash {
         size_t operator()(const Voxel &voxel) const {
             const uint32_t *vec = reinterpret_cast<const uint32_t *>(voxel.data());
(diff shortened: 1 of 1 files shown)


nachovizzo:
Hello, @ybrunken. Thanks again for contributing. We are indeed aware of the non-primeness of the numbers. Indeed @benemer is already updating OpenVDB, where we originally took inspiration for the hashing function at https://github.com/AcademySoftwareFoundation/openvdb/pull/1626

If I may, I would like to ask some questions:
- Could you comment on why solved the hashing function the way you did? 
- Did you experience any improvement at all? After changing the function ourselves we couldn't really see any difference in runtime/accuracy (we also did not instrument the hashing library at all). This question is out of curiosity only

Thanks!

PR #190, opened 2023-07-14
Update boreas.py

kurthobein:
Changed variable name of yaw from 'y' to 'ya'


nachovizzo:
Thanks for catching this one!

PR #193, opened 2023-07-18
Tiziano/normalize timestamps

tizianoGuadagnino:
Add min-max normalization for timestamps on ROS1/ROS2 and python rosbag dataloader

### DIFF
--- a/python/kiss_icp/tools/point_cloud2.py
+++ b/python/kiss_icp/tools/point_cloud2.py
@@ -81,8 +81,10 @@ def read_point_cloud(msg: PointCloud2) -> Tuple[np.ndarray, np.ndarray]:
     points = points[~np.any(np.isnan(points), axis=1)]
 
     if t_field:
-        timestamps = points_structured[t_field]
-        timestamps = timestamps / np.max(timestamps) if t_field != "time" else timestamps
+        timestamps = points_structured[t_field].astype(np.float64)
+        min_timestamp = np.min(timestamps)
+        max_timestamp = np.max(timestamps)
+        timestamps = (timestamps - min_timestamp) / (max_timestamp - min_timestamp)
     else:
         timestamps = np.ones(points.shape[0])
     return points.astype(np.float64), timestamps
--- a/ros/ros1/Utils.hpp
+++ b/ros/ros1/Utils.hpp
@@ -57,9 +57,9 @@ inline auto GetTimestampField(const PointCloud2::ConstPtr msg) {
 
 // Normalize timestamps from 0.0 to 1.0
 inline auto NormalizeTimestamps(const std::vector<double> &timestamps) {
-    const auto min_max_it = std::minmax_element(timestamps.cbegin(), timestamps.cend());
-    const double min_timestamp = *min_max_it.first;
-    const double max_timestamp = *min_max_it.second;
+    const auto [min_it, max_it] = std::minmax_element(timestamps.cbegin(), timestamps.cend());
+    const double min_timestamp = *min_it;
+    const double max_timestamp = *max_it;
 
     std::vector<double> timestamps_normalized(timestamps.size());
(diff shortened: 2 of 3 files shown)


nachovizzo:
please sync with main branch

tizianoGuadagnino:
Done bruda

ros/ros1/Utils.hpp
```diff
@@ -57,12 +57,15 @@ inline auto GetTimestampField(const PointCloud2::ConstPtr msg) {
 
 // Normalize timestamps from 0.0 to 1.0
 inline auto NormalizeTimestamps(const std::vector<double> &timestamps) {
-    const double max_timestamp = *std::max_element(timestamps.cbegin(), timestamps.cend());
-    // check if already normalized
-    if (max_timestamp < 1.0) return timestamps;
+    const auto min_max_it = std::minmax_element(timestamps.cbegin(), timestamps.cend());
```
nachovizzo, line 60: use structured bidnings     `const auto [a,b] = std::minmax_element(timestamps.cbegin(), timestamps.cend());`

ros/ros2/Utils.hpp
```diff
@@ -58,12 +58,15 @@ inline auto GetTimestampField(const PointCloud2::ConstSharedPtr msg) {
 
 // Normalize timestamps from 0.0 to 1.0
 inline auto NormalizeTimestamps(const std::vector<double> &timestamps) {
-    const double max_timestamp = *std::max_element(timestamps.cbegin(), timestamps.cend());
-    // check if already normalized
-    if (max_timestamp < 1.0) return timestamps;
+    const auto min_max_it = std::minmax_element(timestamps.cbegin(), timestamps.cend());
```
nachovizzo, line 61: structured bindings again

python/kiss_icp/tools/point_cloud2.py
```diff
@@ -81,8 +81,10 @@ def read_point_cloud(msg: PointCloud2) -> Tuple[np.ndarray, np.ndarray]:
     points = points[~np.any(np.isnan(points), axis=1)]
 
     if t_field:
-        timestamps = points_structured[t_field]
-        timestamps = timestamps / np.max(timestamps) if t_field != "time" else timestamps
+        timestamps = points_structured[t_field].astype(np.float64)
+        min_timestamp = np.min(timestamps)
+        max_timestamp = np.max(timestamps)
+        timestamps = (timestamps - min_timestamp) / (max_timestamp - min_timestamp)
```
nachovizzo, line 87: same with offset

nachovizzo (changes requested):

ros/ros1/Utils.hpp
tizianoGuadagnino: Fix in new commit

ros/ros1/Utils.hpp
```diff
@@ -57,12 +57,15 @@ inline auto GetTimestampField(const PointCloud2::ConstPtr msg) {
...
+    const double min_timestamp = *min_max_it.first;
+    const double max_timestamp = *min_max_it.second;
+
     std::vector<double> timestamps_normalized(timestamps.size());
     std::transform(timestamps.cbegin(), timestamps.cend(), timestamps_normalized.begin(),
-                   [&](const auto &timestamp) { return timestamp / max_timestamp; });
+                   [&](const auto &timestamp) {
+                       return (timestamp - min_timestamp) / (max_timestamp - min_timestamp);
```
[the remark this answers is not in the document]
tizianoGuadagnino: Timestamps need to be normalized between 0 and 1. Because we select the middle pose in the scan (t=0.5), if min_timestamp is not zero we are taking a random pose in the middle of the the sweep. With this offset we guaranty that the timestamps are always in a valid range. This as not an impact on our pipeline because everything got shifted by the same amount.

nachovizzo (approved):

PR #198, opened 2023-08-01
Nacho/scikit build core

nachovizzo:
Refactor python build system backend. Favor scikit-build-core in favor of scikit-build.

For a list of new features, please check the official [docs](https://scikit-build-core.readthedocs.io/en/latest/index.html)

Besides simplifying the python package, the most "interesting" feature is the `auto rebuilds on import`. Which now it will launch a build when you run the pipeline, if the package was installed as editable (`make editable`)


benemer:
Looks good to me

PR #214, opened 2023-08-29
Path fix, to use dataloader outside kiss-icp

sumanthrao1997:
The tum dataloader works with kiss-icp, but to use it as stand alone dataloader, small fix is needed


sumanthrao1997:
Also made a small change to remove, repeated time stamps in ground truth

PR #218, opened 2023-09-05
Update pipeline.py

kurthobein:
Changed computation of `avg_ms` to use float value of `_get_fps()` instead of int `avg_fps`, which caused `ZeroDivisionError` if avg_fps was evaluated to 0.

### DIFF
--- a/python/kiss_icp/pipeline.py
+++ b/python/kiss_icp/pipeline.py
@@ -178,7 +178,7 @@ class OdometryPipeline:
         avg_tra, avg_rot = sequence_error(self.gt_poses, self.poses)
         ate_rot, ate_trans = absolute_trajectory_error(self.gt_poses, self.poses)
         avg_fps = int(np.floor(_get_fps()))
-        avg_ms = 1e3 * (1 / avg_fps)
+        avg_ms = 1e3 * (1 / _get_fps())
 
         self.results.append(desc="Average Translation Error", units="%", value=avg_tra)
         self.results.append(desc="Average Rotational Error", units="deg/m", value=avg_rot)


python/kiss_icp/pipeline.py
```diff
@@ -178,7 +178,7 @@ def _get_fps():
         avg_tra, avg_rot = sequence_error(self.gt_poses, self.poses)
         ate_rot, ate_trans = absolute_trajectory_error(self.gt_poses, self.poses)
         avg_fps = int(np.floor(_get_fps()))
-        avg_ms = 1e3 * (1 / avg_fps)
+        avg_ms = 1e3 * (1 / _get_fps())
```
nachovizzo, line 181: Wouldn't it be better to replace `np. floor` for `np.ceil` anyways? Having 0fps does not make sense! Thanks for this catch. This is what I would do:

```diff
diff --git a/python/kiss_icp/pipeline.py b/python/kiss_icp/pipeline.py
index e8d0b3a..21d8c66 100644
--- a/python/kiss_icp/pipeline.py
+++ b/python/kiss_icp/pipeline.py
@@ -177,8 +177,8 @@ class OdometryPipeline:
         # Run metrics evaluation
         avg_tra, avg_rot = sequence_error(self.gt_poses, self.poses)
         ate_rot, ate_trans = absolute_trajectory_error(self.gt_poses, self.poses)
-        avg_fps = int(np.floor(_get_fps()))
-        avg_ms = 1e3 * (1 / avg_fps)
+        avg_fps = int(np.ceil(_get_fps()))
+        avg_ms = int(np.ceil(1e3 * (1 / _get_fps())))
 
         self.results.append(desc="Average Translation Error", units="%", value=avg_tra)
         self.results.append(desc="Average Rotational Error", units="deg/m", value=avg_rot)
```

python/kiss_icp/pipeline.py
```diff
@@ -166,24 +166,22 @@ def _write_gt_poses(self):
...
-        avg_tra, avg_rot = sequence_error(self.gt_poses, self.poses)
-        ate_rot, ate_trans = absolute_trajectory_error(self.gt_poses, self.poses)
-        avg_fps = int(np.floor(_get_fps()))
-        avg_ms = 1e3 * (1 / avg_fps)
-
-        self.results.append(desc="Average Translation Error", units="%", value=avg_tra)
-        self.results.append(desc="Average Rotational Error", units="deg/m", value=avg_rot)
-        self.results.append(desc="Absolute Trajectory Error (ATE)", units="m", value=ate_trans)
-        self.results.append(desc="Absolute Rotational Error (ARE)", units="rad", value=ate_rot)
+        avg_fps = int(np.ceil(_get_fps()))
```
nachovizzo, line 183: Taking the opportunity here to always spit the timing results, no matter if GT is available or not

PR #224, opened 2023-09-19
fix kitti rotation error unit

YuePanEdward:
I found that the unit for average rotation error (kitti metric) might be wrong. 
If `deg/m` is used for average rotation error's unit as [here](https://github.com/PRBonn/kiss-icp/blob/d3bb87caf98229b08fe32a7259845237dc51cbc7/python/kiss_icp/pipeline.py#L174):

```
self.results.append(desc="Average Rotational Error", units="deg/m", value=avg_rot)
```

and according to [here](https://github.com/PRBonn/kiss-icp/blob/main/cpp/kiss_icp/metrics/Metrics.cpp#L152):

```
double avg_rot_error = 100.0 * (r_err / static_cast<double>(err.size())) / 3.14 * 180.0;
```

then the calculated value needs to be divided by 100.

### DIFF
--- a/cpp/kiss_icp/metrics/Metrics.cpp
+++ b/cpp/kiss_icp/metrics/Metrics.cpp
@@ -149,7 +149,7 @@ std::tuple<float, float> SeqError(const std::vector<Eigen::Matrix4d> &poses_gt,
     }
 
     double avg_trans_error = 100.0 * (t_err / static_cast<double>(err.size()));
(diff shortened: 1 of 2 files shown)


benemer:
Good catch, thanks! Indeed, the unit of the average relative rotation error is deg/m, so it does not make sense to scale it by a factor of 100. 

However, instead of fixing this in the `metrics.py`, I think it would make more sense to correct it in the `Metrics.cpp` (because it's already wrong here). Could you please change it there??

benemer:
Perfect, thanks!

Issue #237, opened 2023-10-10
reset initial pose through ros service


RamanaBotta:
Hello, 
### About
Since kiss-icp is prone to accumulation of positional errors over time, having something like a service call to reset odometry would really help when working with other sensors, especially sensors like GPS (globally consistent).
### Existing solutions
This kind of feature is available with Zed Cemera positional tracking, Where through a service call it we can reset the start position of odometry [Ref](http://wiki.ros.org/zed-ros-wrapper).

nachovizzo:
I agree this would be great to have. Would you like to contribute it to the project? I can guide you a bit once you have a prototype ready ;)

RamanaBotta:
I am happy to contribute :)

RamanaBotta:
I found a related code snippet in the Zed wrapper.
https://github.com/stereolabs/zed-ros-wrapper/blob/71eb2bb434f059e17191503b707267938f5a1b7f/zed_nodelets/src/zed_nodelet/include/zed_wrapper_nodelet.hpp#L408

RamanaBotta:
Hello, please find the pull request regarding this issue.

PR #241, opened 2023-10-12
feat: added Reset initial pose through ros service

RamanaBotta:
## Issue
https://github.com/PRBonn/kiss-icp/issues/237 
## What?
kiss-icp is not consistent for global localization, so we need to have a reset_pose feature when working with GPS or other global localization systems.
So this pull request adds such feature to provide such a feature to the ros1 version of kiss-icp.

## Why?
Since kiss-icp is prone to accumulation of positional errors over time, having something like a service call to reset odometry would really help when working with other sensors, especially sensors like GPS (globally consistent).

## How?
- Created a service file in ros.
- Added a New method, reset_pose  at  kiss_icp::pipeline::Reset [Here](https://github.com/RamanaBotta/kiss-icp/blob/8f681536505da3a024ef632c0fe9001ad2f45b24/cpp/kiss_icp/pipeline/KissICP.hpp#L70)
- And calling this method once service is called from ros.

## Testing?

1. ```roslaunch kiss_icp odometry.launch```
2. ```rosservice call /reset_pose "{x: 0.0, y: 0.0, z: 0.0, R: 0.0, P: 0.0, Y: 0.0}" ```

#### Repository:
https://github.com/RamanaBotta/kiss-icp
#### Branch to be tested:
Reset-initial-pose-through-ros-service
## Observations
- When /reset_pose service was called, kiss-icp odometry should start from the same position as requested in the service call.
- In My local system, ros-noetic, Ubuntu 20.04, Results are as expected.
## Expected Results
- When /reset_pose service was called, kiss-icp odometry should start from the same position as requested in the service call.
## Screenshots (optional)
Screenshot from 2023-10-12 22-17-06
Screenshot from 2023-10-12 22-16-51


nachovizzo:
Please fix the ci! Install ore commit and run it locally. You can also check the email you got about the error

cpp/kiss_icp/pipeline/KissICP.cpp
```diff
@@ -107,4 +107,11 @@ bool KissICP::HasMoved() {
     return motion > 5.0 * config_.min_motion_th;
 }
 
+void KissICP::Reset(Eigen::Quaterniond quaternion, Eigen::Vector3d translation) {
```
nachovizzo, line 110: This method does not reset the poses but instead injects a new one... I'd split this into two methods. What do you think?

nachovizzo:
@RamanaBotta , I'm adding @benemer and @tizianoGuadagnino to join the review party.

As a general comment, the draft looks ok. But it's far from our standards. KISS-ICP is a multi-framework project. This PR is only affecting the ROS 1 side; what you would be missing is the following: 
- [ ] Implement this in the Python API
- [ ] Implement this in the ROS 2 API

I know it's too much to ask, but we will never merge this in the current state. We spend a lot of time and effort to make the system what it is right now. I hope you understand :smiley: 

What do you think? Would you be able to do the extra effort and complete the integration?

cpp/kiss_icp/pipeline/KissICP.cpp
RamanaBotta: Alright, My understanding from your comment, 
1 . Reset to be done when we Reset to the origin position.
2 . Since This feature is setting the starting position, we need to have a method SetPose.
Is that what you are referring to?

RamanaBotta:
Hello @nachovizzo, Thank you for assigning reviewers. 
@tizianoGuadagnino, @benemer Welcome to this pull request. 

@nachovizzo, I can add support for ROS2 API. But I need some support on Python API.
I am happy to put in extra effort and complete the integration with your support.

cpp/kiss_icp/pipeline/KissICP.hpp
```diff
@@ -67,6 +67,7 @@ class KissICP {
     double GetAdaptiveThreshold();
     Sophus::SE3d GetPredictionModel() const;
     bool HasMoved();
+    void Reset(Eigen::Quaterniond, Eigen::Vector3d);
```
tizianoGuadagnino, line 70: Please use Sophus::SE3d as argument. Quaternion&Vector is more of a ROS thing and I will like to keep the cpp API Sophus only.

cpp/kiss_icp/pipeline/KissICP.cpp
```diff
@@ -107,4 +107,11 @@ bool KissICP::HasMoved() {
...
+    poses_.clear();
+    Sophus::SE3d predpred(quaternion, translation);
+    poses_.push_back(predpred);
+    local_map_.Clear();
```
tizianoGuadagnino, line 114: Instead of resetting the local map here, would in make more sense to transformed it the new reference frame so that KISS can continue his "job" with the old map points?

What do you think ? @nachovizzo @benemer @RamanaBotta

benemer: I agree, we shold transform the local map and the poses to the new frame. Otherwise, every time `Reset` is called, the constant velocity model can not be applied and the map is empty. So you lose a lot of the nice properties of KISS.

RamanaBotta: > Instead of resetting the local map here, would in make more sense to transformed it the new reference frame so that KISS can continue his "job" with the old map points?
> 
> What do you think ? @nachovizzo @benemer @RamanaBotta

I agree, It makes sense to transform the local map to the new reference frame.

cpp/kiss_icp/pipeline/KissICP.hpp
RamanaBotta: Sure.

RamanaBotta: Solved

cpp/kiss_icp/pipeline/KissICP.cpp
RamanaBotta: Hello @tizianoGuadagnino @benemer ,
I have tried transforming the local map to the new reference frame. Please look [here](https://github.com/RamanaBotta/kiss-icp/blob/6f17fcd32cd8c0c2059c386e7db61e3d194723a8/cpp/kiss_icp/pipeline/KissICP.cpp#L110) for implementation.
### Issue 
After resetting to the new frame, the Local map is not looking good, It has previous ref frame points too.
Screenshot from 2023-10-14 00-13-14
### Help
Please consider looking into implementation [here](https://github.com/RamanaBotta/kiss-icp/blob/6f17fcd32cd8c0c2059c386e7db61e3d194723a8/cpp/kiss_icp/pipeline/KissICP.cpp#L110), please let me know any corrections or suggestions.

benemer: Can you please push your changes to this branch? This makes it easier to discuss it.

For example, you are still clearing the poses when resetting the map. This will prevent the system from de-skewing and using a constant velocity model because the velocity can not be estimated without prior pose information. If it is desired to reset the list of estimated poses, I would at least transform and keep the last two.

RamanaBotta: Hi @benemer, Can you Please reiterate this suggestion of transforming the local map, without this also things are working fine.

tizianoGuadagnino: Hi @RamanaBotta, for us it is not just important that it works but it also needs to be correct from the theoretical perspective. Once the pose has been reset, it just makes sense to transform the local map in the new reference frame, so that KISS can use the previous map. Please fix this.

nachovizzo:
Hello, thanks for the amazing work. This branch got a bit outdated and the PR is getting a bit bigger than expected. I'm closing the PR now in favor of opening a new one. But please try to do so once the code has been tested, and it's in a final state. Otherwise, the collaboration from our side turns to be a bit challenging since we none of us is working full-time in this project.

Thanks a lot!
image

RamanaBotta:
Okay @nachovizzo ,
Can you please tell me, How to get support from you without the pull request?

Issue #244, opened 2023-10-17
[ROS 1] Backport tf multi-sensor fix 


nachovizzo:
Apply the changes done for https://github.com/PRBonn/kiss-icp/pull/232 in the ROS 1 wrapper.

It's literally a copy-paste exercise, so shouldn't be that "hard"

PR #245, opened 2023-10-18
Backport tf multisensor fix

playertr:
Resolves https://github.com/PRBonn/kiss-icp/issues/244 : backports the changes to the tf buffer from ROS 2 to ROS 1, with the eventual goal of supporting multi-sensor online odometry.

This PR also changes a few minor bugs in the ROS 2 implementation:
- Removes the default base frame from the launch file, which previously confounded the `egocentric_estimation` computation
- Adds Sophus to the package.xml and CMakelists

Please check:

- [x] Memory correctness: the `unique_ptr` usage and move semantics are less idiomatic to ROS 1 than to ROS 2, but I think my implementation is legit
- [x] Default names: we use the absolute `/kiss` namespace preface for our topics, but it might be more idiomatic to just put these nodes in a Namespace group in the launch file

### DIFF
--- a/ros/CMakeLists.txt
+++ b/ros/CMakeLists.txt
@@ -47,11 +47,13 @@ if("$ENV{ROS_VERSION}" STREQUAL "1")
     COMPONENTS geometry_msgs
                nav_msgs
                sensor_msgs
+               geometry_msgs
                roscpp
                rosbag
                std_msgs
                tf2
-               tf2_ros)
+               tf2_ros
+               Sophus)
   catkin_package()
 
   # ROS 1 node
@@ -71,6 +73,7 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
   find_package(rclcpp_components REQUIRED)
   find_package(sensor_msgs REQUIRED)
   find_package(tf2_ros REQUIRED)
+  find_package(sophus REQUIRED)
 
   # ROS 2 node
   add_library(odometry_component SHARED ros2/OdometryServer.cpp)
@@ -84,7 +87,9 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
     rclcpp_components
     nav_msgs
     sensor_msgs
-    tf2_ros)
+    geometry_msgs
+    tf2_ros
+    Sophus)
 
   rclcpp_components_register_node(odometry_component PLUGIN "kiss_icp_ros::OdometryServer" EXECUTABLE odometry_node)
 
--- a/ros/ros1/OdometryServer.cpp
+++ b/ros/ros1/OdometryServer.cpp
@@ -52,10 +52,9 @@ using utils::PointCloud2ToEigen;
 
 OdometryServer::OdometryServer(const ros::NodeHandle &nh, const ros::NodeHandle &pnh)
     : nh_(nh), pnh_(pnh) {
-    pnh_.param("child_frame", child_frame_, child_frame_);
+    pnh_.param("base_frame", base_frame_, base_frame_);
     pnh_.param("odom_frame", odom_frame_, odom_frame_);
-    pnh_.param("publish_alias_tf", publish_alias_tf_, true);
-    pnh_.param("publish_odom_tf", publish_odom_tf_, true);
+    pnh_.param("publish_odom_tf", publish_odom_tf_, false);
     pnh_.param("max_range", config_.max_range, config_.max_range);
     pnh_.param("min_range", config_.min_range, config_.min_range);
     pnh_.param("deskew", config_.deskew, config_.deskew);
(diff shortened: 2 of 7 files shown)


playertr:
I added this PR! Thank you for contributing the ROS 2 version of this, @nachovizzo . Hope I didn't work redundantly to anyone -- this addition was helpful for my work.

I actually still have a question about the intended usage of the `base_frame`, `odom_frame` , and `publish_odom_tf` parameters (which we should document in the README after sorting out, btw). I noticed that, in the bag I was developing in, the `publish_odom_tf` code was not having the intended effect. 

Even though the transformBroadcaster was sending a transform, that transform was neglected by Rviz, presumably because the sensor frame of my vehicle had _already_ been given a parent by another odometry system. For me, Rviz could not visualize the transform from `kiss/odom` to `velodyne_points_frame`. Given this possibility, should be publish the `odom_frame` as a child of the base link to prevent possible violating tf's DAG assumption?

ros/ros1/OdometryServer.cpp
```diff
@@ -76,101 +75,97 @@ OdometryServer::OdometryServer(const ros::NodeHandle &nh, const ros::NodeHandle
...
     // Initialize publishers
-    odom_publisher_ = pnh_.advertise<nav_msgs::Odometry>("odometry", queue_size_);
-    frame_publisher_ = pnh_.advertise<sensor_msgs::PointCloud2>("frame", queue_size_);
-    kpoints_publisher_ = pnh_.advertise<sensor_msgs::PointCloud2>("keypoints", queue_size_);
-    map_publisher_ = pnh_.advertise<sensor_msgs::PointCloud2>("local_map", queue_size_);
-
-    // Initialize trajectory publisher
+    odom_publisher_ = pnh_.advertise<nav_msgs::Odometry>("/kiss/odometry", queue_size_);
+    frame_publisher_ = pnh_.advertise<sensor_msgs::PointCloud2>("/kiss/frame", queue_size_);
+    kpoints_publisher_ = pnh_.advertise<sensor_msgs::PointCloud2>("/kiss/keypoints", queue_size_);
```
nachovizzo, line 80: I got rid of these debugging publishers in the ROS 2 wrapper. To avoid doing additional transformations just for visualization. At the end, the frame is "identical" (if you don't deksew). And I think the python API is better anyways for such debugging scenarios...

ros/CMakeLists.txt
```diff
@@ -47,11 +47,13 @@ if("$ENV{ROS_VERSION}" STREQUAL "1")
...
                nav_msgs
                sensor_msgs
+               geometry_msgs
                roscpp
                rosbag
                std_msgs
                tf2
-               tf2_ros)
+               tf2_ros
+               Sophus)
```
nachovizzo, line 56: From this one, I'm not quite sure what is the "best" practice... KISS-ICP has its own, isolated and self-contained build system... This means that the only way you couldn't "link" against on of its dependencies is when you copy-paste by hand some source files. 

On the other hand, if we add sophus here, then, we should also add eigen, TBB, etc...  

Do you have a better idea for this?

ros/launch/odometry.launch.py
```diff
@@ -42,7 +42,7 @@ def generate_launch_description():
             DeclareLaunchArgument("bagfile", default_value=""),
             DeclareLaunchArgument("visualize", default_value="true"),
             DeclareLaunchArgument("odom_frame", default_value="odom"),
-            DeclareLaunchArgument("base_frame", default_value="base_footprint"),
+            DeclareLaunchArgument("base_frame", default_value=""),
```
nachovizzo, line 45: I'm unsure which would be the best sane "default" for this one.  I guess most ROS applications would already have a `base_link` or `base_footprint` defined in the tf tree.

@tizianoGuadagnino any comments?

ros/ros1/OdometryServer.hpp
```diff
@@ -42,15 +44,19 @@ class OdometryServer {
...
+    Sophus::SE3d CloudToBaseTf(const std::string &pointcloud_frame_id) const;
+
     /// Ros node stuff
     ros::NodeHandle nh_;
     ros::NodeHandle pnh_;
     int queue_size_{1};
 
     /// Tools for broadcasting TFs.
-    tf2_ros::TransformBroadcaster tf_broadcaster_;
+    std::unique_ptr<tf2_ros::TransformBroadcaster> tf_broadcaster_;
```
nachovizzo, line 56: I'd avoid using pointers here. As far as I know, there is no benefit on ROS 1 (compared to composable nodes in ROS 2). So, this "complicates" the implementation for no good reason (as I see it now).  Again, please challenge any bs**it I say :) Happy to change my mind al;ways

ros/ros1/OdometryServer.cpp
```diff
@@ -76,101 +75,97 @@ OdometryServer::OdometryServer(const ros::NodeHandle &nh, const ros::NodeHandle
...
+}
+
 void OdometryServer::RegisterFrame(const sensor_msgs::PointCloud2::ConstPtr &msg) {
+    const auto cloud_frame_id = msg->header.frame_id;
     const auto points = PointCloud2ToEigen(msg);
     const auto timestamps = [&]() -> std::vector<double> {
         if (!config_.deskew) return {};
         return GetTimestamps(msg);
     }();
+    const auto egocentric_estimation = (base_frame_.empty() || base_frame_ == cloud_frame_id);
```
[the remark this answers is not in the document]
playertr: No, I think that logic is fine. If the base_frame_ is empty or is the same as the cloud_frame_id then we should be in egocentric mode and return our trajectory and map in the cloud frame.

ros/launch/odometry.launch.py
playertr: If we leave this parameter empty by default, then at each iteration `egocentric_estimation` will evaluate to True they way it's written now. As a result, the local map and trajectory will be reported in the sensor frame instead of the base frame, regardless of the specifics of the user's TF tree. If we set the default argument to `base_link`, then I think that attempting to compute `pose` will fail if  `base_link` is not connected to the frame of the input PointCloud2 in the user's TF tree:

```
[ERROR] [1697646211.202267521]: Failed to find tf between base_link and crl_rzr/base. Reason=Could not find a connection between 'velodyne-link' and 'base_link' because they are not part of the same tree.Tf has two or more unconnected trees.
```

Therefore, to me, leaving `base_frame` empty seems like a less limiting option.

ros/ros1/OdometryServer.cpp
playertr: Nice catch! Yes, I accidentally left these publishers in even though they are no longer used.

ros/CMakeLists.txt
playertr: Hmmm, that's a good question. I added the Sophus dependency because I got a linker error for Sophus when doing the following build process for ROS 1:
```
cd ~/catkin_ws/src
git clone https://github.com/PRBonn/kiss-icp
cd ~/catkin_ws
rosdep install --from-paths src --ignore-src -r -y
catkin build
# the compiler initially complained that Sophus could not be found
# but worked once I added the dependency to CMakeLists.txt and package.xml
```

I think that, because we explicitly include Sophus and Eigen headers in the OdometryServer code, we should have a dependency on these in package.xml and CMakeLists.txt. That way, if kiss-icp-cpp ever somehow removed its dependencies on these packages in the future (unlikely) then the ROS packages would still compile. It feels risky to include a dependency's dependency based on prior knowledge of the build process without explicitly adding it to the package.xml.

(I'm keen to know if there's a better way to do this too.)

ros/ros1/OdometryServer.cpp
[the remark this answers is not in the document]
nachovizzo: yes, sorry :)

ros/launch/odometry.launch.py
nachovizzo: Ok, I agree on this one. Will replicate also this on the ROS 2 node

ros/CMakeLists.txt
nachovizzo: Linking error is unlikely since sophus is a header-only library. On the other hand, lets take for example `tf2_ros`, just because is the one above: 
https://github.com/ros/geometry2/blob/309814fd7d55d874f9bac40b2cc8669a5b86f3c5/tf2_ros/CMakeLists.txt#L9

We don't need to include `actionlib` as our dependency. That's the whole magic of build systems.

On the other other hand, I also agree that just because we explicitly include the headers in the Odometryserver, that injects a hard dependency.... so it might be cleaner to explicitly put it there. My problem is with ROS-based build systems, that's if you put sophus on the `package.xml` and it's not in your system, then the build should fail. In contrast, KISS will fetch this at build time...

I'm puzzled about which is the best approach for this case. I will meditate and be back at this PR

nachovizzo:
@playertr would you mind updating this branch with the latest changes form the ros2 driver? Thanks!

And sorry for making you work twice :(

playertr:
> @playertr would you mind updating this branch with the latest changes form the ros2 driver? Thanks!
> 
> And sorry for making you work twice :(

Sure, I'll rebase off of `upstream/nacho/fix_ros_tf_tree_usage` (not `upstream/nacho/fix_ros_tf_tree_usage_clean`).

nachovizzo:
Merging this and finishing some bits at the big PR ;) Thanks a lot for the contribution!

PR #246, opened 2023-10-19
Extract publishing mechanism to a new function

nachovizzo:
Just opening this to review this a bit easier.

The whole idea of this is to reduce the reading burden when reasoning/inspecting the main logic. Publishing is something no we don't really care :) 

Like this, when I look at the main function of the server, I can really see what's going on:
image


nachovizzo:
@benemer happy to revert this, I'm merging this to move forward a bit ;)

benemer (commented):
Approved post-merge :upside_down_face:

PR #265, opened 2024-01-12
Support Pydantic V2

ahans:
This updates the code to be compatible with Pydantic V2. We also introduce a smoke test for importing `kiss_icp`. It's still just a smoke test, but does a bit more than only calling `kiss_icp_pipeline --version`. The latter would also pass with Pydantic V2 being installed, while the import would fail.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -25,3 +25,7 @@ jobs:
       - name: Test installation
         run: |
           kiss_icp_pipeline --version
+      - name: Run unittests
+        run: |
+          python -m pip install --verbose './python[test]'
+          pytest -rA --verbose ./python/
--- a/python/kiss_icp/config/parser.py
+++ b/python/kiss_icp/config/parser.py
@@ -28,7 +28,6 @@ import sys
 from pathlib import Path
 from typing import Any, Dict, Optional
 
-from pydantic import PrivateAttr
 from pydantic_settings import BaseSettings
 
 from kiss_icp.config.config import AdaptiveThresholdConfig, DataConfig, MappingConfig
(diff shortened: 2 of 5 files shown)


ahans:
This isn't quite ready yet.

python/kiss_icp/config/parser.py
```diff
@@ -38,32 +39,23 @@ class KISSConfig(BaseSettings):
     data: DataConfig = DataConfig()
     mapping: MappingConfig = MappingConfig()
     adaptive_threshold: AdaptiveThresholdConfig = AdaptiveThresholdConfig()
-    _config_file: Optional[Path] = PrivateAttr()
-
-    def __init__(self, config_file: Optional[Path] = None, *args, **kwargs):
-        self._config_file = config_file
```
ahans, line 44: Accessing an attribute before the parent's constructor was called is not possible with Pydantic V2. But even with the previous version, the pattern here is not considered very "pydantic" (see, e.g., [here](https://stackoverflow.com/a/65873630/250826)). Having the model class as plain as possible and filling it from the outside seems to be cleaner.

nachovizzo:
@ahans I'm more than happy to see you back in the PR section. Thanks a lot!!! I'm checking this now

.github/workflows/python.yml
```diff
@@ -25,3 +25,7 @@ jobs:
       - name: Test installation
         run: |
           kiss_icp_pipeline --version
+      - name: Run unittests
```
nachovizzo, line 28: I will see this on the CI and pretend we have unit tests 😜 Just won't open the tab

python/kiss_icp/config/parser.py
nachovizzo: Thanks for clarifying this!

nachovizzo (approved):
Great! Happy to upgrade to the new pydantic version! Thanks for leading the effort

Waiting on @benemer for approval + merge

benemer (approved):
All good from my side, thanks for the support :)

ahans:
@nachovizzo thanks for catching my cleanup failures!

Issue #268, opened 2024-01-24
Map becomes invisible in global view


benemer:
When using the global view in the visualizer and the default black background, the local map's x and y coordinates will quickly go out of the bounds of the colormap and will be rendered just black and therefore invisible. This looks like the registration failed and lost the map, but it's just rendering:

![image](https://github.com/PRBonn/kiss-icp/assets/38326482/4a901bd6-a000-4be1-9c58-d93df47204fe)

The easiest solution would be to paint the map when using the global view.

tizianoGuadagnino:
Solved in #269

PR #269, opened 2024-01-24
Paint map in global view

benemer:
Solves https://github.com/PRBonn/kiss-icp/issues/268

When using global view, we now paint the map in gray (which can be seen with a black and white background). Even though the default coordinate-based rendering looks better for local map inspection, it will paint the local map black in the global view. In the local view (which should be used to inspect the map) we keep the original rendering.

image

I am also open to other color suggestions or ways to rescale the default colormap :)


nachovizzo (approved):
LGTM!

tizianoGuadagnino (approved):

Issue #259, opened 2023-11-23
Robustness under frame rate changes


zyxu1996:
I used a 30-second 10HZ point cloud and randomly dropped 20 frames of it. I noticed that pose's estimates were not ideal, and the point clouds did not match up well on discrete times. Is there any way to adjust the parameters in config to deal with this? Or do I have to use an IMU to optimize it?

nachovizzo:
Would you mind explaining the meaning of:

> I used a 30-second 10HZ point cloud and randomly dropped 20 frames of it.

zyxu1996:
There's a point cloud with a timestamp like this

zyxu1996:
There's a discontinuous point cloud with a timestamp like this (s means second):
0.1s, 0.2s, 0.3s, 0.5s, 0.7s, 0.8s, 0.9s, 1.0s, 1.4s, 1.5s, 1.6s
It lost 0.4s, 0.6s, 1.1s, 1.2s, 1.3s. It will most likely give a wrong pose prediction at 1.0s or 1.4s.

zyxu1996:
> Would you mind explaining the meaning of:
> 
> > I used a 30-second 10HZ point cloud and randomly dropped 20 frames of it.

There's a discontinuous point cloud with a timestamp like this (s means second):
0.1s, 0.2s, 0.3s, 0.5s, 0.7s, 0.8s, 0.9s, 1.0s, 1.4s, 1.5s, 1.6s
It lost 0.4s, 0.6s, 1.1s, 1.2s, 1.3s. It will most likely give a wrong pose prediction at 1.0s or 1.4s.

benemer:
Our [prediction](https://github.com/PRBonn/kiss-icp/blob/3465687027471419acb1e425d2e22ca02a360925/python/kiss_icp/kiss_icp.py#L87) model

```python
    def get_prediction_model(self):
        if len(self.poses) < 2:
            return np.eye(4)
        return np.linalg.inv(self.poses[-2]) @ self.poses[-1]
```
assumes that the scans are equidistant in time (therefore we can use the delta between the last two estimated poses as the initial guess for the next one). This does not hold in your case, so you will sometimes use the pose change within 0.2 seconds as an initial guess for a movement within 0.1 seconds (which is wrong).

The first thing you should try is taking the actual time difference between estimations into account to predict an initial guess.

tizianoGuadagnino:
[...]

PR #270, opened 2024-01-27
Improve frame drop robustness

yhao10:
For real time performance, we may face lidar frame drop cases like the https://github.com/PRBonn/kiss-icp/issues/259 issue.

My chance (for python part only) allows user to pass in the dropped frame number to the algorithm to get a better initial position estimation and a more accurate deskew results.

The main API is self.kiss_icp.register_frame(xyz, ts, frame_delta). frame_delta here is the number of dropped frames of the current register frame. Users can obtain it from the lidar frame ID or adding a frame counter in their code.

### DIFF
--- a/cpp/kiss_icp/core/Deskew.cpp
+++ b/cpp/kiss_icp/core/Deskew.cpp
@@ -38,11 +38,8 @@ std::vector<Eigen::Vector3d> DeSkewScan(const std::vector<Eigen::Vector3d> &fram
                                         const std::vector<double> &timestamps,
                                         const Sophus::SE3d &start_pose,
                                         const Sophus::SE3d &finish_pose,
-                                        int prev_frame_delta) {
-    if (prev_frame_delta <= 0) {
-	throw std::runtime_error("Math error. prev_frame_delta has to be greater than zero.");
-    }
-    const auto delta_pose = (start_pose.inverse() * finish_pose).log() / prev_frame_delta;
+					double frame_delta_ratio) {
+    const auto delta_pose = (start_pose.inverse() * finish_pose).log() * frame_delta_ratio;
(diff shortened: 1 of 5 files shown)


yhao10:
Two tests I did for these changes.

1. Constant frame drop: drop rate is 1 frame for every 4 frames
    Original code process time is 191.1978 seconds vs new change process time is 179.7402 seconds
    Original code drift (due to frame drop): Total travelled 626.505 m and average diff 9.878 m. Off around 1.577%
    while new change drift: INFO:root:Total travelled 626.505 m and average diff 9.552 m. Off around 1.525%

2. Random frame drop: drop percent is 25% and maximum 3 continuous frames drop
    Original code process time is 204.7789 seconds vs new change process time is 230.3906 seconds
    Original code drift: Total travelled 627.115 m and average diff 10.403 m. Off around 1.659%
    while new change drift: Total travelled 625.745 m and average diff 8.480 m. Off around 1.355%

I see shorter process time (ICP converge quicker) and a little bit less drift.

benemer:
Thank you for your contribution to KISS-ICP!

However, there are some issues with this PR. First, as Tiziano already pointed out in the referenced issue:

> KISS is not meant to handle specific hardware issues such as frame dropping. This probably requires some robustification of the pipeline to wrong initial guesses, which is not the goal of KISS

Since frame dropping is a rather specific hardware issue, we do not want to explicitly address it with KISS because it will make the project more complicated without needing it in most cases.

Besides that, your solution is not ideal. The frame drop ratio is a rather unintuitive parameter that has to be computed by the user. It would be better to read (if available) the timestamps of the scans in the dataloader and then use these for correctly deskewing the frame and predicting the motion. This would approach the problem in a more general way which in my view is more KISS-like :)

Best
Benedikt

PR #272, opened 2024-01-30
Fix incorrect installation hyperlink

kerryeon:
Hello, thanks for operating great works!

This PR fixes an incorrect link in `README.md`.

If a guest user clicks the link below,:

https://github.com/PRBonn/kiss-icp/blob/656bcaeddfb819fa89c32e6c44666d58a3ad6972/README.md?plain=1#L12

then the user has no permission to view, because it's editor mode.

If a normal user clicks the link, then the user would watch the installation markdown script, not the compiled web content.

So, this PR fixes the link from `editor mode` to `viewer mode` like the others.


benemer (approved):
Good catch, thanks!

PR #275, opened 2024-02-11
Refactor getClosestNeighbour

jtaveau:
I took the initiative to refactor this function as it's the most expensive one of the project (computationnally speaking). 
The results looked great on hotspot, but I had only a small bag of 15s to test it, would you have a bigger one? 

Before refactor:
Screenshot from 2024-02-11 13-26-14

After refactor:
Screenshot from 2024-02-11 13-08-32

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -49,47 +49,37 @@ namespace kiss_icp {
 VoxelHashMap::Vector3dVectorTuple VoxelHashMap::GetCorrespondences(
     const Vector3dVector &points, double max_correspondance_distance) const {
     // Lambda Function to obtain the KNN of one point, maybe refactor
-    auto GetClosestNeighboor = [&](const Eigen::Vector3d &point) {
+    auto GetClosestNeighbour = [&](const Eigen::Vector3d &point) {
         auto kx = static_cast<int>(point[0] / voxel_size_);
         auto ky = static_cast<int>(point[1] / voxel_size_);
         auto kz = static_cast<int>(point[2] / voxel_size_);
-        std::vector<Voxel> voxels;
-        voxels.reserve(27);
+
+        Eigen::Vector3d closest_neighbour;
+        double closest_distance2 = std::numeric_limits<double>::max();
+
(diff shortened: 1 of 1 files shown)


nachovizzo:
@jtaveau thanks a lot for the initiative. Indeed the hot path from the code is looking for nearest neighbors. As it has been the case for most systems out there as well

I'll review the code and test this intensively, with all the datasets we know were it's working and come back to you. It might take some time only

PS: thanks for fixing the typos 😂

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -49,64 +49,54 @@ namespace kiss_icp {
...
-        Vector3dVector neighboors;
-        neighboors.reserve(27 * max_points_per_voxel_);
-        std::for_each(voxels.cbegin(), voxels.cend(), [&](const auto &voxel) {
-            auto search = map_.find(voxel);
-            if (search != map_.end()) {
-                const auto &points = search->second.points;
-                if (!points.empty()) {
-                    for (const auto &point : points) {
-                        neighboors.emplace_back(point);
+                    auto search = map_.find({i, j, k});
```
nachovizzo, line 63: Would you mind being a bit more explicit (on the PR description, not the code) about the change you just made?

I kind of reverse engineered the new solution and I think is better, but if would be nice for the other reviewers to facilitate a bit the high level idea ;)

nachovizzo (commented):
Looks good to me! I tried on a few datasets and it's working, although I do not see any real improvement in terms of speed.

I think I prefer the other implementation since the 3 tasks that the lambda function is doing where defined in 3 different steps. With this implementation they all tree tasks got merged into quite a few nested loops, which complicates redability for new users

I will let Tiziano and Benedikt comment and test on this particular PR

Great job @jtaveau

cpp/kiss_icp/core/VoxelHashMap.cpp
jtaveau: I agree the readability is not great, I'll profile again the code this evening to see if there is some real improvements and give explanations

nachovizzo:
Closing this to avoid stalling, feel free to re-open when you manage to benchmark it!

Best!

PR #281, opened 2024-02-29
Remove always True flag

nachovizzo:
This also makes the C++ and Python pipelines closer, since C++ has no option to enable/disable preprocessing

I have **no strong** opinions, so we can decide whether to keep this config. I haven't realized that we never use it, as "preprocessing" is just cropping the min/max range of the cloud, so I'd say it is something we always want to do....

Let me know what you guys think

### DIFF
--- a/config/advanced.yaml
+++ b/config/advanced.yaml
@@ -5,7 +5,6 @@ out_dir: "results"
 
 data:
   deskew: True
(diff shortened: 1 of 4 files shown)


tizianoGuadagnino (approved):
I love it <3

PR #282, opened 2024-02-29
Refactor ROS parametrization

nachovizzo:
Expose KISS parameters in the launch file leaving the ROS arguments exposed through the launch configuration

~Besides being a more standard and "ROS"y practice, this will allow us in the future for a more flexible configuration of the ROS NODE. The number of parameters on the ROS side of things is likely to grow, and maintaining this in the launch file will become a mess.~

~This also lets users have a more flexible usage of the ROS node for experimentation, one can now select at launch time with a parameter file to use. In fact, many people are [already doing this](https://github.com/giafranchini/etna_s3li_playback/blob/main/config/kiss_icp.yaml)~

**Important **: This PR does not change any piece of the implementation, it's just a cosmetic change on how we manipulate the ROS wiring, nothing else


tizianoGuadagnino:
I am not super excited about this; it goes a bit against our initial philosophy of separating stuff. I will instead have a slightly larger launchfile that merges the two configs. As we know, the parameters of the base pipeline usually work without intervention, so it makes sense that the use of "tweak" is just the ros part in a specific launch file. Also I think it gives more flexibility for developers working on non-ROS framework (if any). As always we can discuss about it

nachovizzo:
> I am not super excited about this; it goes a bit against our initial philosophy of separating stuff. I will instead have a slightly larger launchfile that merges the two configs. As we know, the parameters of the base pipeline usually work without intervention, so it makes sense that the use of "tweak" is just the ros part in a specific launch file. Also I think it gives more flexibility for developers working on non-ROS framework (if any). As always we can discuss about it

I'm also not super as well, let me see if I can improve the launch.py to merge the configs there

nachovizzo:
@tizianoGuadagnino please re-review.

Now I merged the YAML file into the launch.py. The only parameters that are exposed to the ROS CLI are:
```
base_frame = LaunchConfiguration("base_frame", default="")
odom_frame = LaunchConfiguration("odom_frame", default="odom")
publish_odom_tf = LaunchConfiguration("publish_odom_tf", default=True)
pointcloud_topic = LaunchConfiguration("topic")
visualize = LaunchConfiguration("visualize", default="true")
bagfile = LaunchConfiguration("bagfile", default="")
```
Meaning that any of the above lists can be modified in the launch file but also when invoking it with the cli.

This way KISS remains configured by default, and from outside we can wire different frames for the tf tree. **NOTE** that this means that this is not possible anymore (but I guess it never made sense anyway)
```sh
ros2 launch kiss_icp odometry.launch.py initial_threshold:=0.4
```


If one needs further customization then one can create a custom launch file to launch the node with the required config:
```py
from launch.substitutions import LaunchConfiguration
from launch_ros.actions import Node


def generate_launch_description():
    return LaunchDescription(
        [
            Node(
                package="kiss_icp",
                executable="kiss_icp_node",
                name="kiss_icp_node",
                output="screen",
                remappings=[("pointcloud_topic", LaunchConfiguration("topic"))],
                parameters=[
                    {
                        "publish_odom_tf": False,
                        "max_range": 20.0,
                    },
                ],
            )
        ]
    )
```

tizianoGuadagnino:
Mmmm ok, let me drive it to see how it feels ;)

tizianoGuadagnino:
I got some weird core dump but I guess I'm doing something wrong. The default Python pipeline kind of dies. I will investigate why, but maybe also @benemer can have a look.

image

nachovizzo:
> I got some weird core dump but I guess I'm doing something wrong. The default Python pipeline kind of dies. I will investigate why, but maybe also @benemer can have a look.
> 
> image

But I'm not modifying the Python pipeline in this PR :thinking: , maybe is this the new pydantic thing?

tizianoGuadagnino:
This is kind of dangerous; it can fire back very easily. We should probably handle this in a separate PR. Me and @benemer will have a look.

tizianoGuadagnino:
Now back to your thing @nachovizzo

nachovizzo:
@benemer , @tizianoGuadagnino fix is already on its way !

train of PRs.... :bullettrain_front: 

https://github.com/PRBonn/kiss-icp/pull/286

raw-t (approved):
Looks good to me :)

benemer (approved):
To me as well!

PR #283, opened 2024-02-29
Add odometry covariance to ros node

nachovizzo:
Depends: https://github.com/PRBonn/kiss-icp/pull/282

Adding a default value for the covariance matrix in the odometry msg.

Far from being ideal, at least now users can tweak this value. If we later realize that it's better to have a full matrix, like in [other open source systems](https://github.com/cra-ros-pkg/robot_localization/blob/d37b824b7ff6e08183801c4d8b58bd41c08f94f0/params/ekf.yaml#L227-L241) then we might upgrade later. For now, I put the same value on the translation/rotational part. The user can only modify 3 values at the same time (Cxx == Cyy == Czz), and (Crr, Cpp, Cyawyaw) accordingly.

Note: By briefly inspecting the current forks of KISS-ICP, most of the cases people are just forking to add this exact bit


nachovizzo:
@tizianoGuadagnino I think this would be amazing. Yes.

The question is, would it make sense to have the 2 options:

- a. automatically compute covariance from the linear system
- b. let the user hardcode covariance values

benemer:
I agree with @tizianoGuadagnino to compute some covariance estimate.

@nachovizzo do you mean having the computed covariances by default and optionally you can set them yourself? I favor this over having hardcoded default values because otherwise, the user might wonder how they are picked.

nachovizzo:
> I agree with @tizianoGuadagnino to compute some covariance estimate.
> 
> @nachovizzo do you mean having the computed covariances by default and optionally you can set them yourself? I favor this over having hardcoded default values because otherwise, the user might wonder how they are picked.

Yes, similar to the adaptive threshold configuration. Something like (pseudo-code)
```
# position_covariance = 0.1 # <- automatically determined by KISS-ICP, uncomment to force a hardcoded value
# same for orientation ..
```

nachovizzo:
In any case, I was thinking that we can merge this (after reviewing dependant PR) to move forward and open a ticket for improving this in the future. I believe that even having a hardcoded value is better than not having anything, as this allow to quickly play around with packages like `r_l`

benemer (approved):
Okay, if you say say it's beneficial then let's merge this with fixed covariances and open an new issue for computing the covariance.

nachovizzo:
added a ticket so once we merge this we don't let the auto-covariance computation fly away from our minds https://github.com/PRBonn/kiss-icp/issues/296

tizianoGuadagnino (approved):

PR #284, opened 2024-02-29
Remove unnecessary path publishing pipeline

nachovizzo:
We can still visualize the trajectory with the odometry msg. It makes no sense from my point of view to overload the system publishing a full path just for visualization. 

_NOTE_ : This is a tech-debt PR, no rush for merging it

### DIFF
--- a/ros/rviz/kiss_icp.rviz
+++ b/ros/rviz/kiss_icp.rviz
@@ -174,34 +174,6 @@ Visualization Manager:
             Reliability Policy: Best Effort
             Value: /kiss/odometry
           Value: true
-        - Alpha: 1
-          Buffer Length: 1
-          Class: rviz_default_plugins/Path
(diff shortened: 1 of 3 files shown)


benemer (approved):

nachovizzo:
The alternative option would be to add a separate python node that subscribes to the odometry-published by kiss and publishes the trajectory, just for debugging purposes. I don't want to have this enabled by default, that's it.

benemer:
I am fine with removing it

tizianoGuadagnino (approved):
I honestly always found it useless, so I am totally fine with this.

nachovizzo:
> I honestly always found it useless, so I am totally fine with this.

Mee too, one day I'll add a silly node in the python launcher, so it doesn't mess with the C++ Node, it's just for seeing a trajectory..  but adds nothing but noise now

PR #285, opened 2024-02-29
Fix and improve ROS visualization

nachovizzo:
First of all, now I can finally enjoy reading the implementation. Without the spaghetti if/else/if/else trying to guess which transformation should be applied.

My approach here was to replicate the same as we do in the Python visualizer.:

https://github.com/PRBonn/kiss-icp/blob/44d8563a992b037f3607d328b8815bf41e029e23/python/kiss_icp/tools/visualizer.py#L223

The new logic is as follows:

-  When debugging using ROS, everything happens in an ego-centric world, from the sensor's point of view
- Since the KISS-ICP local map is on the odometric coordinate frame, we need to apply the last ego-centric pose to bring it back to the lidar frame.
- This allowed for further simplification

The only caveat is that, if we DO NOT PUBLISH to the tf tree, then **only** the trajectory will be on a different reference frame and would look odd in the viz. The rest is still fine. Since this is something we never used, need, we don't have in Python, I figured out that's worth disabling it by default.
image

The rest I think is all improvement so far

**NOTE** I haven't changed the implementation of KISS, just the visualization utilities for ROS

### DIFF
--- a/python/kiss_icp/tools/visualizer.py
+++ b/python/kiss_icp/tools/visualizer.py
@@ -29,7 +29,7 @@ from typing import Callable, List
 
 import numpy as np
 
-YELLOW = np.array([1, 0.706, 0])
+CYAN = np.array([0.24, 0.898, 1])
 RED = np.array([128, 0, 0]) / 255.0
 BLACK = np.array([0, 0, 0]) / 255.0
 BLUE = np.array([0.4, 0.5, 0.9])
@@ -198,7 +198,7 @@ class RegistrationVisualizer(StubVisualizer):
         # Source hot frame
         if self.render_source:
             self.source.points = self.o3d.utility.Vector3dVector(source)
-            self.source.paint_uniform_color(YELLOW)
+            self.source.paint_uniform_color(CYAN)
             if self.global_view:
                 self.source.transform(pose)
         else:
@@ -207,7 +207,7 @@ class RegistrationVisualizer(StubVisualizer):
         # Keypoints
         if self.render_keypoints:
             self.keypoints.points = self.o3d.utility.Vector3dVector(keypoints)
-            self.keypoints.paint_uniform_color(YELLOW)
+            self.keypoints.paint_uniform_color(CYAN)
             if self.global_view:
                 self.keypoints.transform(pose)
         else:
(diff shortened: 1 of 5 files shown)


benemer:
> Since this is something we never used, need, we don't have in Python, I figured out that's worth disabling it by default.

But showing the point clouds and the full trajectory is something we can show in the Python visualizer or I am getting it wrong? Or are you talking about the default in the visualizer?

nachovizzo:
> > Since this is something we never used, need, we don't have in Python, I figured out that's worth disabling it by default.
> 
> But showing the point clouds and the full trajectory is something we can show in the Python visualizer or I am getting it wrong? Or are you talking about the default in the visualizer?

You are right, it's something we can show on the Python side. Here it's always populating this infinitely wrong path which I don't think it has an application on the ROS side of things. The path can also be seen in RVIZ by just clicking in the odometry topic, which we don't have in Python

nachovizzo:
> @nachovizzo as a general comment not strictly related to the issue (which looks good as far as I saw), would it make sense to always publish the tf (a.k.a. we remove the `publich_odom_tf_` flag) ?

@tizianoGuadagnino Sadly no. No because it doesn't make sense, but rather due to a limitation of tf. Assume the transformation you want to publish is already present in the tree, and then publishing it from Kiss will create some conflicts.

I always wonder if we should go back to the hacky "alias" tf we had, just to publish a completely disconnected tree to make sure we control that.

On the other hand, we also have a strong limitation with rviz. Since, the visualization frame can't be variable, and thus is hardcoded. I'd prefer to do like we do in the python visualizer and see everything from the lidar eyes, but rviz forbid us to do so

tizianoGuadagnino:
Ok, now it is clear why, I didn't think about that. 

I have mixed feelings about the alias, although I really like the idea of having a bit more control on the tf tree, even if it requires some hacks. Maybe it is something worth discussing offline with @benemer .

Anyways, from my side, we can merge this.

tizianoGuadagnino (approved):

nachovizzo:
I also have mixed feelings, especially because it adds a ton of code that is virtually necessary. But all these problems are to avoid a user launching the pipeline and seeing a black screen because we couldn't guess :crystal_ball: the `frame_id` from who knows which lidar is being used.

Complicated tasks to tackle in my opinion if we want to stay generic and don't force the user to read 5 pages of documentation on how to configure rviz

benemer (approved):

PR #286, opened 2024-03-01
Prepend `kiss_icp_` for env variable usage

nachovizzo:
Before this didn't work
```
export DATA=carlos
kiss_icp_pipeline ...
```

After, you need to be more explicit to break it:
```
export KISS_ICP_DATA=carlos
kiss_icp_pipeline ...
```

This allows us to maintain the API as well

### DIFF
--- a/python/kiss_icp/config/parser.py
+++ b/python/kiss_icp/config/parser.py
@@ -28,12 +28,13 @@ import sys
 from pathlib import Path
 from typing import Any, Dict, Optional
 
-from pydantic_settings import BaseSettings
+from pydantic_settings import BaseSettings, SettingsConfigDict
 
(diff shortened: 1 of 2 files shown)


benemer:
Awesome! This solves it. I added some information to the help, I think otherwise no one will realize that this is possible. What do you think @nachovizzo ?

nachovizzo:
> Awesome! This solves it. I added some information to the help, I think otherwise no one will realize that this is possible. What do you think @nachovizzo ?

looks great, thanks

benemer (approved):
Perfect, then let's go :bullettrain_side:

tizianoGuadagnino:
Love you guys <3

tizianoGuadagnino (approved):

PR #289, opened 2024-03-01
Remove unnecessary exposed utility function from ROS API

nachovizzo:
Remove unnecessary exposed utility function from ROS API

With this change, this function is not exposed (which was never the
intention to) to the header. This way, we can "hide" this into private unnamed namespaces and benefit from inlining the simple function into the translation unit.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @nachovizzo @benemer @tizianoGuadagnino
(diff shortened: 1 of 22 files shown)


benemer (approved):

tizianoGuadagnino (dismissed):

tizianoGuadagnino:
Fix the merge confict ;)

PR #292, opened 2024-03-07
Expose number of icp iterations in ros

nachovizzo:
Draft PR, depends on #282, #283, #252


nachovizzo:
Auto mergint this since it's just exposing more params to launch files

PR #294, opened 2024-03-11
Auto 2257 fix double construction kiss icp cpp

nachovizzo:
_NOTE:_ I sneaked some small changes alongside this PR to avoid stacking 20 other PRs. I tried to be more verbose commit-wise (don't just review the full diff) and also provided a description here

Testing https://github.com/PRBonn/kiss-icp/pull/252 I realized that there was a small buggy problem in the construction of the C++ pipeline:

Since we were using a stack KISS-ICP pipeline, we were constructing twice, once at compile time with default parameters and 2nd at runtime with the runtime ROS params

Besides this being ugly led to a horrible bug, that luckily I captured early, which was the fact that I couldn’t control the number of threads (static const g) since it was default constructed to multithreaded and ignored at runtime :man_facepalming: 

## Changes

- Delete `kiss_icp::pipeline::KissICP` default constructor  15e20a222fc2d469f1db838f3c1f7. Python system is not affected, nothing is affected
- Update the only consumer, the ROS wrapper:  20c2d7419a7eb
- Rename `odometry_` way too generic 8c38132bc14741f4e27709b32931cad4c3bd5f

## Minor changes

Since I was forced to review the horror I wrote last year, I also:
- Remove a duplicated code 3fed3a60049124a3ff6b1fe6c5 
- Removed `config_` to be an unnecessary class member of the odometry server 250708a61db1388, and update the utils accordingly at 533047cbd13a1856c5e3a5a2

### DIFF
--- a/cpp/kiss_icp/pipeline/KissICP.cpp
+++ b/cpp/kiss_icp/pipeline/KissICP.cpp
@@ -37,7 +37,7 @@ namespace kiss_icp::pipeline {
 KissICP::Vector3dVectorTuple KissICP::RegisterFrame(const std::vector<Eigen::Vector3d> &frame,
                                                     const std::vector<double> &timestamps) {
     const auto &deskew_frame = [&]() -> std::vector<Eigen::Vector3d> {
-        if (!config_.deskew) return frame;
+        if (!config_.deskew || timestamps.empty()) return frame;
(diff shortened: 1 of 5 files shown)


benemer (approved):
Good catch, looks good from my side!

tizianoGuadagnino (approved):
I love this one, awesome bruda ;)

PR #300, opened 2024-03-24
Fix bug where point would match with random point

02alexander:
When `GetClosestNeighbor` can't find any point in it's neighborhood it returned an uninitialized vector which caused nonsense correspondences. Sometimes these correspondences get filtered out by the adaptive threshold by not always, below is a screenshot of a case when it's not filtered out.
kiss-icp-bug
In the screenshot the red points are from the local map, the blue points are from the `source` and the green lines are the matches from the ICP loop.

### DIFF
--- a/cpp/kiss_icp/core/Registration.cpp
+++ b/cpp/kiss_icp/core/Registration.cpp
@@ -65,7 +65,7 @@ std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels = 1
     return voxel_neighborhood;
 }
 
-std::tuple<double, Eigen::Vector3d> GetClosestNeighbor(const Eigen::Vector3d &point,
+std::tuple<Eigen::Vector3d, double> GetClosestNeighbor(const Eigen::Vector3d &point,
                                                        const kiss_icp::VoxelHashMap &voxel_map) {
     // Convert the point to voxel coordinates
     const auto &voxel = voxel_map.PointToVoxel(point);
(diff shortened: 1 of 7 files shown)


raw-t:
Hi @02alexander, thanks for pointing this out. I believe the optional is unnecessary in this case, as we are computing the distance to the closest neighbor into the `GetClosestNeighbor` and outside. We could just use this value to filter out uninitialized points, I'm working on it now.

raw-t:
@02alexander I just push on your repository, I hope it is fine otherwise fill free to revert ;) Tell me if you like the change, and thanks for spotting this.

What do you think guys? @nachovizzo @benemer

benemer (approved):
@02alexander great catch! And thank you for the contribution.

Looks good to me! Now we even skip the second distance computation :)

nachovizzo:
Guys this is amazing!!!!!! thanks a lot, @02alexander for catching this, and thanks @tizianoGuadagnino for kissifying the proposal. I also like the fact we got rid of a duplicated norm computation

nachovizzo (approved):

tizianoGuadagnino (dismissed):

nachovizzo (approved):

tizianoGuadagnino (approved):

benemer (approved):

02alexander:
@tizianoGuadagnino, I like your solution of returning the distance instead of an optional.

02alexander:
@nachovizzo, here is an video of the ICP with the commited changes. It no more random correspondences. 

https://github.com/PRBonn/kiss-icp/assets/28707703/d10d8aa9-9cae-48aa-be6e-570c490eaa5e

tizianoGuadagnino:
@02alexander perfect, thanks again for your contribution :)

PR #303, opened 2024-03-26
Use string for sequence

benemer:
I would like to change the sequence parameter to a `str` instead of an `int`. This allows more flexibility when, for example, using a dataset like HeliPR (dataloader coming soon):

image

Here, there are four different sensors available for each recording. With this, we could pass the desired sensor as the sequence parameter:

```
kiss_icp_pipeline --dataloader helipr /path/to/HeLiPR/Roundabout01 -v -s Aeva
```

Passing a KITTI sequence ID will still work (and is anyway internally converted to a string):

```
kiss_icp_pipeline --dataloader kitti /path/to/kitti-odometry/dataset -s 0
```

Otherwise, we have to pass the full path `/path/to/HeLiPR/Roundabout01/LiDAR/Aeva` as `data_dir`, and this becomes messy since the reference poses the dataloader has to read in a parent directory in `/path/to/HeLiPR/Roundabout01/LiDAR_GT`.

What do you guys think @nachovizzo @tizianoGuadagnino?

### DIFF
--- a/python/kiss_icp/datasets/kitti.py
+++ b/python/kiss_icp/datasets/kitti.py
@@ -28,7 +28,7 @@ import numpy as np
 
 class KITTIOdometryDataset:
     def __init__(self, data_dir, sequence: str, *_, **__):
-        self.sequence_id = sequence.zfill(2)
+        self.sequence_id = str(sequence).zfill(2)
         self.kitti_sequence_dir = os.path.join(data_dir, "sequences", self.sequence_id)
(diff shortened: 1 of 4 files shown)


tizianoGuadagnino (commented):
I like it, I think it will generalize the sequence param. We basically use it just for KITTI while it can be useful also in other places where the sequence is not an int.

nachovizzo (approved):

tizianoGuadagnino (approved):

PR #305, opened 2024-03-27
Change default to None

benemer:
Since the CLI arguments for `deskew` and `max_range` overwrite the config and the default argument for `deskew` is `False`, the `deskew` parameter from the config was never used.


tizianoGuadagnino (approved):
Thank you Benedetto :heart:

nachovizzo (approved):

PR #307, opened 2024-03-28
Add more timestamps

benemer:
For Apollo, we can switch to using Open3D's tensor library to read the additional attributes from the `pcd.` Note that Apollo is already de-skewed, so it's not needed, but it's nice to have access to it.

NCLT uses a Velodyne-32, so I used the same naive timestamp extraction based on the range image columns as in KITTI raw.

This btw requires Open3d >= 0.16, but I think that should not be a problem.

### DIFF
--- a/python/kiss_icp/datasets/apollo.py
+++ b/python/kiss_icp/datasets/apollo.py
@@ -49,11 +49,14 @@ class ApolloDataset:
         return len(self.scan_files)
 
     def __getitem__(self, idx):
-        return self.get_scan(self.scan_files[idx])
+        return self.read_point_cloud(self.scan_files[idx])
 
-    def get_scan(self, scan_file: str):
-        points = np.asarray(self.o3d.io.read_point_cloud(scan_file).points, dtype=np.float64)
-        return points.astype(np.float64)
+    def read_point_cloud(self, scan_file: str):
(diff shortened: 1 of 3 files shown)


benemer:
For Apollo, this makes the pipeline a bit slower, unsure if it is because of using `o3d.t`. Maybe double check on your side. We can also remove the timestamps again if we don't want to sacrifice runtime.

nachovizzo:
> For Apollo, this makes the pipeline a bit slower, unsure if it is because of using `o3d.t`. Maybe double check on your side. We can also remove the timestamps again if we don't want to sacrifice runtime.

:thinking: then no need to add it right ? or what would be the reasoning on adding the timestamps if we can't use it in the pipeline

benemer:
Actually no. It was just something I did before realizing it was not needed for KISS-ICP. If people use our dataloaders for other projects it might be interesting, but I am fine with removing it.

nachovizzo:
then I guess the addition for NCLT is nice

benemer:
I also checked again the NCLT dataset, and here they say for the `velodyne_sync` directory:

> these points are motion compensated for the Segway’s ego-motion during
the scan and then recorded in the Segway body frame

Since we decided only to put timestamps if they are used by our pipeline, we can actually close this PR :)

Sorry for the noise!

PR #310, opened 2024-04-01
VoxelHashMap ereasing of the voxels

### LINKED ISSUE #309: VoxelHashMap ereasing of the voxels
**First of all, I would love to thank you for your brilliant work and elegant solution.**
I would like to propose change of one thing from code safety perspective that could potentially lead to undefined behavior.

## Description
I've encountered an issue with erasing voxels in VoxelHashMap. The problem is as follows:

- Current implementation is using **tsl::robin_map**, which is Robin Hood hashing with backward shift deletion. 
- This means that when an element is removed from a tsl::robin_map, the elements are shifted backward to fill the gap left by the removed element.
- Removing elements from a container while iterating over it can result in unexpected behavior, such as skipping elements or processing the same element multiple times. 
- It's generally considered unsafe practice to modify a container while iterating over it, regardless of the specific container type.

 I've proposed solution to this problem.

### Solution:
- This version is more safe and won't result in any undefined behavior.
- The drawback is: It is less elegant solution and harder to read.

## Linked Commits
- [The proposed solution](https://github.com/kulfer80/kiss-icp/commit/f95142d0abd09c553987d07756ff11a7cfa3267f)

## Testing
I've tested proposed solution locally and confirmed that they don't alter the behavior of the pipeline.

Please read the proposed solution and feel free to provide any feedback. Thanks one more time for great work!

kulfer80:
#### The reason for this pull request is to consider the safety of the code so that it is not susceptible to potential unexpected behavior.

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -89,7 +89,7 @@ void VoxelHashMap::RemovePointsFarFromLocation(const Eigen::Vector3d &origin) {
         const auto &[voxel, voxel_block] = *it;
         const auto &pt = voxel_block.points.front();
         const auto max_distance2 = max_distance_ * max_distance_;
-        if ((pt - origin).squaredNorm() > (max_distance2)) {
+        if ((pt - origin).squaredNorm() >= (max_distance2)) {
             it = map_.erase(it);
         } else {
             ++it;
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -88,7 +88,8 @@ void VoxelHashMap::RemovePointsFarFromLocation(const Eigen::Vector3d &origin) {
     for (auto it = map_.begin(); it != map_.end();) {
         const auto &[voxel, voxel_block] = *it;
         const auto &pt = voxel_block.points.front();
-        if ((pt - origin).squaredNorm() > (max_distance2_)) {
+        const auto max_distance2 = max_distance_ * max_distance_;
+        if ((pt - origin).squaredNorm() > (max_distance2)) {
             it = map_.erase(it);
         } else {
             ++it;
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -85,11 +85,13 @@ void VoxelHashMap::AddPoints(const std::vector<Eigen::Vector3d> &points) {
 }
 
 void VoxelHashMap::RemovePointsFarFromLocation(const Eigen::Vector3d &origin) {
-    for (const auto &[voxel, voxel_block] : map_) {
+    for (auto it = map_.begin(); it != map_.end();) {
+        const auto &[voxel, voxel_block] = *it;
         const auto &pt = voxel_block.points.front();
-        const auto max_distance2 = max_distance_ * max_distance_;
-        if ((pt - origin).squaredNorm() > (max_distance2)) {
-            map_.erase(voxel);
+        if ((pt - origin).squaredNorm() > (max_distance2_)) {
+            it = map_.erase(it);
+        } else {
+            ++it;
         }
     }
 }
--- a/cpp/kiss_icp/core/VoxelHashMap.hpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.hpp
@@ -52,7 +52,7 @@ struct VoxelHashMap {
 
     explicit VoxelHashMap(double voxel_size, double max_distance, int max_points_per_voxel)
         : voxel_size_(voxel_size),
-          max_distance_(max_distance),
+          max_distance2_(max_distance * max_distance),
           max_points_per_voxel_(max_points_per_voxel) {}
 
     inline void Clear() { map_.clear(); }
@@ -70,7 +70,7 @@ struct VoxelHashMap {
     std::vector<Eigen::Vector3d> GetPoints(const std::vector<Voxel> &query_voxels) const;
 
     double voxel_size_;
-    double max_distance_;
+    double max_distance2_;
     int max_points_per_voxel_;
     tsl::robin_map<Voxel, VoxelBlock, VoxelHash> map_;
 };
--- a/cpp/kiss_icp/core/VoxelHashMap.hpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.hpp
@@ -52,7 +52,7 @@ struct VoxelHashMap {
 
     explicit VoxelHashMap(double voxel_size, double max_distance, int max_points_per_voxel)
         : voxel_size_(voxel_size),
-          max_distance2_(max_distance * max_distance),
+          max_distance_(max_distance),
           max_points_per_voxel_(max_points_per_voxel) {}
 
     inline void Clear() { map_.clear(); }
@@ -70,7 +70,7 @@ struct VoxelHashMap {
     std::vector<Eigen::Vector3d> GetPoints(const std::vector<Voxel> &query_voxels) const;
 
     double voxel_size_;
-    double max_distance2_;
+    double max_distance_;
     int max_points_per_voxel_;
     tsl::robin_map<Voxel, VoxelBlock, VoxelHash> map_;
 };


kulfer80:
@nachovizzo I did more testing and investigations, this is a disassembled code:
https://godbolt.org/z/GrhMve1eh

---

In this scenario it can be clearly seen that by original solution the one key,value pair isn't deleted that should have been deleted by condition.

kulfer80:
@nachovizzo I have just recently tested both implementations on KITTI_raw dataset with robin_map and in some cases voxels were skipped and they weren't removed, in most of the cases size of map remained same in both implementations. 
@nachovizzo Should I send you some logs or testing scenarios ? Let me know.

kulfer80:
> ```c++
> for (auto it = map_.begin(); it != map_.end();) {
>     if ((pt - origin).squaredNorm() > (max_distance2)) {
>         map_.erase(it);
>     } 
>     ++it;
> }

In this version, what is happening is:
When an element is erased, the elements after it might get shifted, but the iterator keeps moving forward, this can lead to either skipped element or revisiting already processed element.

nachovizzo:
@kulfer80 yes and no :)

Although over the surface looks like what's happening, and indeed makes more sense after reading the documentation I still believe this is not the case. What I think it's true is that after `map_.erase(it)` that particular iterator is no longer valid, `it` points to an element that no longer exists. But since we do not use it again, it does not  change the results (to my eyes). After the `++it` then the iterator will point to the next valid element in the hash table and thus, valid again. 

I would agree that this is kind of "hacky" since we rely on not using such iterator in between the loop, although with the C++17 structured bindings used in the original implementation that iterator is only visible to the compiler and not to the user. So it's should be safe.

I did this very dirty program:
```cpp
#include <iostream>
#include <tsl/robin_map.h>

int main() {
  tsl::robin_map<int, int> map = {
      {1, 10},
      {2, 20},
      {3, 30},
  };
  std::cout << "-------------------------------------" << std::endl;
  for (auto it = map.begin(); it != map.end();) {
    const auto &[key, value] = *it;

    if (value > 10) {
      std::cout << "value before " << std::get<1>(*it) << std::endl;
      std::cout << "size  before " << map.size() << std::endl;
      map.erase(it);
      std::cout << "value  after " << std::get<1>(*it) << std::endl;
      std::cout << "size   after " << map.size() << std::endl;
    }
    ++it; // here the iterator gets updated and no longer invalid
    if (value > 10) {
      std::cout << "value    end " << std::get<1>(*it) << std::endl;
      std::cout << "size     end " << map.size() << std::endl;
      std::cout << "-------------------------------------" << std::endl;
    }
  }
  return 0;
}
```

And this is the result, that somehow supports my shady theory. After erasing the iterator is invalid, but after incrementing it everything goes back to normal

image

Do you have any solid tests that confirm this is a real problem?

Thanks for the help bwt!

kulfer80:
@nachovizzo You are right about validating iterator. The thing is, the original solution that is now implemented in the code would work if every voxel had unique hash, but it some cases it isn't true. 

---

Imagine this situation you have robin_map that contains key, value pairs, in which key == value so simplicity.
The hash is key % 2, so there will be collisions. Now let s say condition is value > 5, we are currently at element with value 6, iterator is pointing at bucket let s say 0x555555573074, the thing that happens after erasing is the element with value 8 is moved to this bucket and we moved iterator to bucket 0x555555573080 so basically we skipped bucket that is occupied by element with value 8 and we finish iterations without removing it, even though condition says it should be removed.

The proposed solution handles it because it returns the same bucket if it is valid iterator if not it will advance to next element.

Here is the snippet that you can try on your own (I know it s ugly written):

```#include <iostream>
#include <tsl/robin_map.h>

struct MyHash {
    std::size_t operator()(int key) const {
        return key % 2;
    }
};

std::pair<int, int> original(int num) {
    tsl::robin_map<int, int, MyHash> map_;
    for (int i = 0; i < 13; ++i) {
        map_[i] = i;
    }

    int count = 0;
    for (auto it = map_.begin(); it != map_.end();) {
        const auto& [key, value] = *it;
        if (value > num) {
            map_.erase(it);
        }
        ++it;
        ++count;
    }

    return std::make_pair(map_.size(), count);
}

std::pair<int, int> proposed(int num) {
    tsl::robin_map<int, int, MyHash> map_;
    for (int i = 0; i < 13; ++i) {
        map_[i] = i;
    }

    int count = 0;
    for (auto it = map_.begin(); it != map_.end();) {
        const auto& [key, value] = *it;
        if (value > num) {
            it = map_.erase(it);
        } else {
            ++it;
        }
        ++count;
    }

    return std::make_pair(map_.size(), count);
}

int main() {
    const auto& [orig_size, orig_count] = original(5);
    const auto& [prop_size, prop_count] = proposed(5);
    std::cout << "Original size: " << orig_size << " count: " << orig_count << std::endl;
    std::cout << "Proposed: size: " << prop_size << " count: " << prop_count<< std::endl;
    return 0;
}

**Comment by nachovizzo** [comment]
@kulfer80 :clap: 
Thanks for keep pushing!

I think now you are right :) So I'm happy to say I'd like to include this.

I don't think we are hitting this problem so many times, and I also don't think this will make any real impact on the system, since we are always working at the boundaries of the voxel map, and I never saw a "non removed" voxel block left behind in visualization

Nevertheless, I do value correct code. So let's move forward with it

The only thing I want to ask is to remove the `max_distance_2`. Just because it's unrelated to this change, after that, ready to go!

File: cpp/kiss_icp/core/VoxelHashMap.cpp [hunk 1]
```diff
@@ -85,11 +85,13 @@ void VoxelHashMap::AddPoints(const std::vector<Eigen::Vector3d> &points) {
...
 
 void VoxelHashMap::RemovePointsFarFromLocation(const Eigen::Vector3d &origin) {
-    for (const auto &[voxel, voxel_block] : map_) {
+    for (auto it = map_.begin(); it != map_.end();) {
+        const auto &[voxel, voxel_block] = *it;
         const auto &pt = voxel_block.points.front();
-        const auto max_distance2 = max_distance_ * max_distance_;
-        if ((pt - origin).squaredNorm() > (max_distance2)) {
-            map_.erase(voxel);
+        if ((pt - origin).squaredNorm() > (max_distance2_)) {
```
nachovizzo, line 91: let's also put a `>=` just to avoid that small corner case where it's for whatever reason exactly the same as `max_distance` and we don't remove the voxel nor forward the iterator

nachovizzo (approved):

nachovizzo:
> @kulfer80 👏 Thanks for keep pushing!
> 
> I think now you are right :) So I'm happy to say I'd like to include this.
> 
> I don't think we are hitting this problem so many times, and I also don't think this will make any real impact on the system, since we are always working at the boundaries of the voxel map, and I never saw a "non removed" voxel block left behind in visualization
> 
> Nevertheless, I do value correct code. So let's move forward with it
> 
> The only thing I want to ask is to remove the `max_distance_2`. Just because it's unrelated to this change, after that, ready to go!

@kulfer80 I did it myself, sorry :roll_eyes: 

Now waiting for the other maintainers review.

@tizianoGuadagnino , @benemer this PR is ready to review. Please take a deep look and give us your opinion :)

kulfer80:
@nachovizzo It really occurred seldom, but if in future you change your mind and replace container robin_map with other it could lead to unexpected behavior. I only created max_distance2_ to not reintizialinzing this variable in every loop cycle. But you are right that way as you rewrote it, it is more readable. And thank you that we could have elaborated about this problem. One more time big thanks to your whole team that did outstanding work 👏.

nachovizzo:
> @nachovizzo It really occurred seldom, but if in future you change your mind and replace container robin_map with other it could lead to unexpected behavior. I only created max_distance2_ to not reintizialinzing this variable in every loop cycle. But you are right that way as you rewrote it, it is more readable. And thank you that we could have elaborated about this problem. One more time big thanks to your whole team that did outstanding work 👏.

feel free to open a 2nd PR about the max_distance. We could check, but I'd assume the compiler is optimizing that bit anyways

kulfer80:
> > @nachovizzo It really occurred seldom, but if in future you change your mind and replace container robin_map with other it could lead to unexpected behavior. I only created max_distance2_ to not reintizialinzing this variable in every loop cycle. But you are right that way as you rewrote it, it is more readable. And thank you that we could have elaborated about this problem. One more time big thanks to your whole team that did outstanding work 👏.
> 
> feel free to open a 2nd PR about the max_distance. We could check, but I'd assume the compiler is optimizing that bit anyways

You are right I overthought it, I disassembled code and it gave the same result in both situations.

tizianoGuadagnino (approved):

nachovizzo:
@kulfer80 congrats on your first contribution :partying_face: !!

PR #312, opened 2024-04-03
[WIP] Covariance Estimation

tizianoGuadagnino:
I introduce the covariance estimation in the KISS pipeline (python/cpp/ROS).

The main change is the introduction of a `struct Estimate` to contain both the pose and the system's covariance for a certain scan. Thanks to some operator overloading and some basic `pybind`, this change should be "invisible" outside of the `Registration.hpp/cpp` files. 

The code is still not as clean as I would like it to be, but I think it is time for @nachovizzo and @benemer to come into the game as well as I work out all the math.

Other than that, give this a look. I will ask you, @nachovizzo, to intensively test this on ROS2 data, as I think you have tons of that.

I have a small mental bug regarding the math, as the covariance will continue to increase as the run goes further, I am studying the problem to see if there is a way to "regularize" this, I will keep you updated.

Let me know what do you think.


ros/src/OdometryServer.hpp
```diff
@@ -79,10 +81,6 @@ class OdometryServer : public rclcpp::Node {
     /// Global/map coordinate frame.
     std::string odom_frame_{"odom"};
     std::string base_frame_{};
-
-    /// Covariance diagonal
```
nachovizzo, line 83: Not something to tackle right away.  But I'd prefer if we keep this fixed covariance as an optional parameter. Similar to how we do (only in python) for the adaptive threshold. I believe many people would like to fuse the KISS ICP output and just give it a constant "weight"

tizianoGuadagnino: Make sense ;)

cpp/kiss_icp/core/Registration.cpp
```diff
@@ -174,35 +198,38 @@ Registration::Registration(int max_num_iteration, double convergence_criterion,
...
-        const Eigen::Vector6d dx = JTJ.ldlt().solve(-JTr);
+        const auto &[JTJ, JTr, chi_square] = BuildLinearSystem(associations, kernel_threshold);
+        const auto JTJ_inverse = JTJ.inverse();
+        const Eigen::Vector6d dx = -JTJ_inverse * JTr;
         const Sophus::SE3d estimation = Sophus::SE3d::exp(dx);
+        icp_correction.pose = estimation * icp_correction.pose;
+        // Follow Ollila et.al. https://arxiv.org/pdf/2006.10005.pdf - Eq. (2) for sample covariance
+        // using an M estimator
+        icp_correction.covariance = 1.0 / static_cast<double>(associations.size()) * JTJ_inverse;
         // Equation (12)
```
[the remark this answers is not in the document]
tizianoGuadagnino: I wanted to keep it there until we don't decide to merge as a reference. I don't want this on `main`, but for now I think it is useful during the development if some of us wants to go a bit deeper in the computation.

cpp/kiss_icp/core/Registration.hpp
```diff
@@ -28,16 +28,27 @@
 
 #include "VoxelHashMap.hpp"
 
+using CovarianceMatrixType = Eigen::Matrix<double, 6, 6>;
+
 namespace kiss_icp {
 
+struct Estimate {
```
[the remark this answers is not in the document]
tizianoGuadagnino: The only alternative I could think about was `PoseWithCovariance` but I have a childhood trauma with that name ...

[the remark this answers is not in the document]
tizianoGuadagnino: But I am open to change it to something of this kind if we all agree

cpp/kiss_icp/core/Registration.cpp
```diff
@@ -131,38 +134,59 @@ LinearSystem BuildLinearSystem(const Associations &associations, double kernel)
...
+// Equations from figure 3 of https://arxiv.org/pdf/1906.07795.pdf
+Estimate Estimate::inverse() const {
+    const auto Adj = pose.inverse().Adj();
+    return {pose.inverse(), Adj * covariance * Adj.transpose()};
+}
+
+Estimate operator*(Estimate lhs, const Estimate &rhs) {
+    lhs.pose *= rhs.pose;
+    const auto Adj = lhs.pose.Adj();
+    lhs.covariance += Adj * rhs.covariance * Adj.transpose();
```
benemer, line 186: In equation 3 of the referenced paper, the right-hand `Adj` is not transposed, is that correct?

tizianoGuadagnino: Yeah, there is a mistake in the figure, but it is correct in the text as I remember. I still reference this paper because the only `open access` that I could find. In general, it is a linear propagation of the covariance, so it should be transpose.

benemer: Ah yes, Eqn. 39. It also makes more sense, as you said :+1:

cpp/kiss_icp/pipeline/KissICP.cpp
```diff
@@ -94,16 +93,19 @@ double KissICP::GetAdaptiveThreshold() {
     return adaptive_threshold_.ComputeThreshold();
 }
 
-Sophus::SE3d KissICP::GetPredictionModel() const {
-    Sophus::SE3d pred = Sophus::SE3d();
-    const size_t N = poses_.size();
-    if (N < 2) return pred;
-    return poses_[N - 2].inverse() * poses_[N - 1];
+Estimate KissICP::GetPredictionModel() const {
+    Estimate prediction;
```
benemer, line 97: I think renaming `Estimate` will also help here, because ` Estimate prediction` looks a bit weird

python/kiss_icp/kiss_icp.py
```diff
@@ -56,21 +56,22 @@ def register_frame(self, frame, timestamps):
 
         # Compute initial_guess for ICP
         prediction = self.get_prediction_model()
-        last_pose = self.poses[-1] if self.poses else np.eye(4)
-        initial_guess = last_pose @ prediction
+        initial_guess = self.estimates[-1] if self.estimates else Estimate()
```
benemer, line 59: Maybe better `last estimate`? Since it is not yet the initial guess

Myzhar:
Hi Guys, any news on this?

tizianoGuadagnino:
> Hi Guys, any news on this?

Unfortunately not, I don't feel we are ready for this change, as in general, I am not convinced that the benefits of adding this "uncertainty" estimation (notice the quotes) are worth this change; this will require more discussions with the KISS  core team, so for now we postponed this at the moment.

vinnnyr:
>  in general, I am not convinced that the benefits of adding this "uncertainty" estimation (notice the quotes) are worth this change; this will require more discussions with the KISS core team, so for now we postponed this at the moment.

I am not necessarily disagreeing, (nor am I agreeing), but if you have a chance, can you please share some of the concerns and critiques of this approach? It seems like maybe I missed something.

PR #314, opened 2024-04-05
Make rosbag dataloader take path to dir

benemer:
We can now pass a path to a directory containing multiple `.bag` files or ROS 2 files with a `metadata.yaml` while specifying the rosbag dataloader. So far, this has only been possible by not passing a dataloader argument and relying on the `guess_dataloader` function to read the bagfiles and turn them into a list of file paths.


tizianoGuadagnino (approved):
Benedetto tu sei Benedetto

nachovizzo (approved):

PR #318, opened 2024-04-08
Refactor KissICP pipeline to mantain just 1 pose instead of full trajectory. Fixes #122

### LINKED ISSUE #122: poses vector grows indefinitely

nachovizzo:
The idea of this PR is to provide a solution to #122 that also improves the whole KISS-ICP implementation

## General Idea
 I think I'm happy in general with moving away to maintaining an infinite list of poses just to always use the current one (`current_pose_`) and the difference to the last (`current_delta_`). So I think the system is benefiting from this change as we don't add extra stuff just because. It models better the problem we are trying to solve and expresses more clearly what we need to keep track of to solve that problem with our approach

## What we changed
-  Removed the `std::vector<Sophus::SE3d>` poses_ member from the C++ pipeline. 
- Added a `current_pose_` (name to be defined) and `current_delta_` (as this is the only variable we were generating from the `poses_[N-2]`
- Extended the deskew function to account for being called with the `delta` estimation instead of the start/end poses. ~(this was not overloaded in the Python module)~  done in a8c39f6a22d709423d2ce12b15564bb6282197f4
- Simplified the adaptive threshold, and removed a bunch of ugly `HasMoved`, etc utility functions
- Try to merge more conceptually the Python and C++ pipelines
- Added some minor comments into the main loops to improve readability for new users

## Important
The Python pipeline will be now the only system that contains the full trajectory. As the C++ one is mainly consumed in ROS it's trivial to create a trajectory by just subscribing to the `kiss/odometry` topic in general

Update: a8c39f6a22d709423d2ce12b15564bb6282197f4 shows how this trajectory should be handled in the "application" and not in the KISS ICP mehotd itself, making the kiss_icp.py even more similar to the C++ evil twin

## To do

- [x] Tests this in a bunch of datasets
- [x] Analyze the there is a real impact with the new approach for the adaptive threshold (97% sure that there is no real change)
- [x] Define what to do with the Deskew API (1 function with just the delta as input, 2 functions, pybinds)
- [x] **DEFINE NEW NAMES**. `delta` Is a bit generic, `current_pose` is the current, the last, which pose? I put `pose` since the moment you ask Kiss for this data member will give you the current estimate it has. Should `delta` be renamed to `prediction` ? I dunno

### DIFF
--- a/cpp/kiss_icp/core/Registration.cpp
+++ b/cpp/kiss_icp/core/Registration.cpp
@@ -41,7 +41,8 @@ using Matrix6d = Eigen::Matrix<double, 6, 6>;
 using Matrix3_6d = Eigen::Matrix<double, 3, 6>;
 using Vector6d = Eigen::Matrix<double, 6, 1>;
 }  // namespace Eigen
-using Associations = std::vector<std::pair<Eigen::Vector3d, Eigen::Vector3d>>;
+
+using Correspondences = std::vector<std::pair<Eigen::Vector3d, Eigen::Vector3d>>;
 using LinearSystem = std::pair<Eigen::Matrix6d, Eigen::Vector6d>;
 
 namespace {
@@ -87,19 +88,19 @@ std::tuple<Eigen::Vector3d, double> GetClosestNeighbor(const Eigen::Vector3d &po
     return std::make_tuple(closest_neighbor, closest_distance);
 }
 
-Associations FindAssociations(const std::vector<Eigen::Vector3d> &points,
-                              const kiss_icp::VoxelHashMap &voxel_map,
-                              double max_correspondance_distance) {
+Correspondences DataAssociation(const std::vector<Eigen::Vector3d> &points,
+                                const kiss_icp::VoxelHashMap &voxel_map,
+                                double max_correspondance_distance) {
     using points_iterator = std::vector<Eigen::Vector3d>::const_iterator;
-    Associations associations;
-    associations.reserve(points.size());
-    associations = tbb::parallel_reduce(
+    Correspondences correspondences;
+    correspondences.reserve(points.size());
+    correspondences = tbb::parallel_reduce(
         // Range
         tbb::blocked_range<points_iterator>{points.cbegin(), points.cend()},
         // Identity
-        associations,
+        correspondences,
         // 1st lambda: Parallel computation
-        [&](const tbb::blocked_range<points_iterator> &r, Associations res) -> Associations {
+        [&](const tbb::blocked_range<points_iterator> &r, Correspondences res) -> Correspondences {
             res.reserve(r.size());
             std::for_each(r.begin(), r.end(), [&](const auto &point) {
                 const auto &[closest_neighbor, distance] = GetClosestNeighbor(point, voxel_map);
@@ -110,17 +111,17 @@ Associations FindAssociations(const std::vector<Eigen::Vector3d> &points,
             return res;
         },
         // 2nd lambda: Parallel reduction
-        [](Associations a, const Associations &b) -> Associations {
+        [](Correspondences a, const Correspondences &b) -> Correspondences {
             a.insert(a.end(),                              //
                      std::make_move_iterator(b.cbegin()),  //
                      std::make_move_iterator(b.cend()));
             return a;
         });
 
-    return associations;
+    return correspondences;
 }
 
-LinearSystem BuildLinearSystem(const Associations &associations, double kernel) {
+LinearSystem BuildLinearSystem(const Correspondences &correspondences, double kernel) {
     auto compute_jacobian_and_residual = [](auto association) {
         const auto &[source, target] = association;
         const Eigen::Vector3d residual = source - target;
@@ -138,14 +139,15 @@ LinearSystem BuildLinearSystem(const Associations &associations, double kernel) 
 
     auto GM_weight = [&](double residual2) { return square(kernel) / square(kernel + residual2); };
 
-    using associations_iterator = Associations::const_iterator;
+    using correspondence_iterator = Correspondences::const_iterator;
     const auto &[JTJ, JTr] = tbb::parallel_reduce(
         // Range
-        tbb::blocked_range<associations_iterator>{associations.cbegin(), associations.cend()},
+        tbb::blocked_range<correspondence_iterator>{correspondences.cbegin(),
+                                                    correspondences.cend()},
         // Identity
         LinearSystem(Eigen::Matrix6d::Zero(), Eigen::Vector6d::Zero()),
         // 1st Lambda: Parallel computation
-        [&](const tbb::blocked_range<associations_iterator> &r, LinearSystem J) -> LinearSystem {
+        [&](const tbb::blocked_range<correspondence_iterator> &r, LinearSystem J) -> LinearSystem {
             return std::transform_reduce(
                 r.begin(), r.end(), J, sum_linear_systems, [&](const auto &association) {
                     const auto &[J_r, residual] = compute_jacobian_and_residual(association);
@@ -177,8 +179,8 @@ Registration::Registration(int max_num_iteration, double convergence_criterion, 
 Sophus::SE3d Registration::AlignPointsToMap(const std::vector<Eigen::Vector3d> &frame,
                                             const VoxelHashMap &voxel_map,
                                             const Sophus::SE3d &initial_guess,
-                                            double max_correspondence_distance,
-                                            double kernel) {
+                                            double max_distance,
+                                            double kernel_scale) {
     if (voxel_map.Empty()) return initial_guess;
 
     // Equation (9)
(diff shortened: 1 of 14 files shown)


nachovizzo:
@tizianoGuadagnino @benemer this is ready for a real round of reviews.

I modified the PR description, title, etc so please re-read.

cpp/kiss_icp/pipeline/KissICP.cpp
```diff
@@ -59,24 +49,32 @@ KissICP::Vector3dVectorTuple KissICP::RegisterFrame(const std::vector<Eigen::Vec
...
-    const auto last_pose = !poses_.empty() ? poses_.back() : Sophus::SE3d();
-    const auto initial_guess = last_pose * prediction;
-
-    // Run icp
-    const Sophus::SE3d new_pose = registration_.AlignPointsToMap(source,         //
-                                                                 local_map_,     //
-                                                                 initial_guess,  //
-                                                                 3.0 * sigma,    //
-                                                                 sigma / 3.0);
+    const auto initial_guess = current_pose_ * current_delta_;
```
nachovizzo, line 56: This only works the very first iteration as `Sophus::SE3d` is default constructed to Identity... maybe a bit fishy but well, worst case we can be more verebsoe in the .hpp

benemer: I am fine with explicitly initializing with identity. Also helps the reading flow if someone is not aware of this in Sophus.

cpp/kiss_icp/core/Threshold.hpp
```diff
@@ -28,9 +28,10 @@ namespace kiss_icp {
 
 struct AdaptiveThreshold {
     explicit AdaptiveThreshold(double initial_threshold, double min_motion_th, double max_range)
-        : initial_threshold_(initial_threshold),
-          min_motion_th_(min_motion_th),
-          max_range_(max_range) {}
+        : min_motion_th_(min_motion_th),
+          max_range_(max_range),
+          model_error_sse2_(initial_threshold * initial_threshold),
```
benemer, line 33: I like simplifying the code by doing this, but I think this is too hard to read/understand. Because we use the initial threshold to compute a hypothetical initial model error (without having computed a single pose, so what does a model error mean here?), such that the threshold matches the initial threshold we want.

However, I can not think of another solution right now that keeps this simplicity.

tizianoGuadagnino: Yeah I done that because in this way we not only have the initial threshold that we want but also the threshold is more robust to the initial _small motion_ issue that we have, for which, at the time, I design the `hasMoved` function. Cannot think about a better way as well, for me we can also hide this in the cpp if that make us feel more in peace.

cpp/kiss_icp/pipeline/KissICP.cpp
```diff
@@ -59,24 +49,32 @@ KissICP::Vector3dVectorTuple KissICP::RegisterFrame(const std::vector<Eigen::Vec
...
+
+    // Save current pose before the ICP loop to compute the current_delta_ later
+    const auto last_pose = current_pose_;
```
benemer, line 59: Is this needed? `current_pose_` should not be changed until line 75 and before we already update `current_delta_`

python/kiss_icp/kiss_icp.py
```diff
@@ -51,13 +53,14 @@ def register_frame(self, frame, timestamps):
...
+        sigma = self.adaptive_threshold.get_threshold()
 
         # Compute initial_guess for ICP
-        prediction = self.get_prediction_model()
-        last_pose = self.poses[-1] if self.poses else np.eye(4)
-        initial_guess = last_pose @ prediction
+        initial_guess = self.current_pose @ self.current_delta
+
+        # Save current pose before the ICP loop to compute the current_delta_ later
+        last_pose = self.current_pose
```
benemer, line 62: This should also be unnecessary. If it's just about avoiding conflicts with current/new naming, we can also rename `self.current_pose` to `self.last_pose` because the current pose is also the last one we computed.

benemer:
Regarding the naming, I would vote for `last_pose_`; this also avoids confusion with the `new_pose`. Regarding the `current_delta_`, how about connecting it directly to the velocity? In this way, it becomes clear that de-skewing is integrating the velocity and that the initial guess is based on the constant velocity model.

benemer:
Regarding Deskewing, I think we can just keep one with the delta/velocity and compute it on the Python side in the MotionCompensator when calling `deskew_scan`

cpp/kiss_icp/pipeline/KissICP.cpp
tizianoGuadagnino: I Agree you benedetto on this, I am now renaming everything from `current_<things>` to `last_<things>`

python/kiss_icp/kiss_icp.py
tizianoGuadagnino: Same as https://github.com/PRBonn/kiss-icp/pull/318#discussion_r1560595602

tizianoGuadagnino:
> Regarding the naming, I would vote for `last_pose_`; this also avoids confusion with the `new_pose`. Regarding the `current_delta_`, how about connecting it directly to the velocity? In this way, it becomes clear that de-skewing is integrating the velocity and that the initial guess is based on the constant velocity model.

Executed now what was current became last, and I also feel now the flow it is more clear

cpp/kiss_icp/core/Registration.cpp
```diff
@@ -87,19 +88,19 @@ std::tuple<Eigen::Vector3d, double> GetClosestNeighbor(const Eigen::Vector3d &po
     return std::make_tuple(closest_neighbor, closest_distance);
 }
 
-Associations FindAssociations(const std::vector<Eigen::Vector3d> &points,
-                              const kiss_icp::VoxelHashMap &voxel_map,
-                              double max_correspondance_distance) {
+Correspondences DataAssociation(const std::vector<Eigen::Vector3d> &points,
```
tizianoGuadagnino, line 91: I did some renaming that was pending from a previous PR, because we are now doing a massive renaming anyway I thought it was the right time. have a look

tizianoGuadagnino:
> Regarding Deskewing, I think we can just keep one with the delta/velocity and compute it on the Python side in the MotionCompensator when calling `deskew_scan`

Didn't fully get what you mean here.

cpp/kiss_icp/core/Registration.cpp
benemer: Looks good to me!

cpp/kiss_icp/core/Threshold.cpp
```diff
@@ -25,28 +25,29 @@
...
+AdaptiveThreshold::AdaptiveThreshold(double initial_threshold,
+                                     double min_motion_threshold,
+                                     double max_range)
+    : min_motion_threshold_(min_motion_threshold),
+      max_range_(max_range),
+      model_sse_(initial_threshold * initial_threshold),
+      model_error_(0.0),
+      num_samples_(1) {}
+
+void AdaptiveThreshold::UpdateModelDeviation(const Sophus::SE3d &current_deviation) {
```
tizianoGuadagnino, line 37: I also did some renaming and reshaping in the `AdaptiveThreshold`, which should improve the understanding. Please @nachovizzo @benemer have a look.

cpp/kiss_icp/core/Threshold.cpp
```diff
@@ -25,28 +25,26 @@
...
+    : min_motion_threshold_(min_motion_threshold),
+      max_range_(max_range),
+      model_sse_(initial_threshold * initial_threshold),
+      num_samples_(1) {}
 
-double AdaptiveThreshold::ComputeThreshold() {
-    double model_error = ComputeModelError(model_deviation_, max_range_);
-    if (model_error > min_motion_th_) {
-        model_error_sse2_ += model_error * model_error;
+void AdaptiveThreshold::UpdateModelDeviation(const Sophus::SE3d &current_deviation) {
```
tizianoGuadagnino, line 37: I also did some renaming and reshaping in the `AdaptiveThreshold`, which should improve the understanding. Please @nachovizzo @benemer have a look.

benemer:
> > Regarding Deskewing, I think we can just keep one with the delta/velocity and compute it on the Python side in the MotionCompensator when calling `deskew_scan`
> 
> Didn't fully get what you mean here @benemer

I was referring to this Todo list's point: 

>  Define what to do with the Deskew API (1 function with just the delta as input, 2 functions, pybinds)

I now removed the deskewing that takes start and end pose because it was not used anyway.

cpp/kiss_icp/core/Threshold.cpp
benemer: Looks good to me :)

benemer (dismissed):

benemer:
Given that the performance is not affected (Tiziano and I will check later), I am now happy with the changes!

nachovizzo:
While I review this I'll split the diffs in different PRs to keep a cleaner changelog and for future criminology purposes :) great work guys thanks!

tizianoGuadagnino:
@nachovizzo I think the log is quite clear, we wanted to use a single pose rather then a vector -> Consequence we need to change the `AdaptiveThreshold` (more specifically we need to handle the initialization phase differently) -> Consequence, we need to evaluate the change in performance -> Consequence we review the metrics.

Any other change is collateral and minor, from my point of view. I would rather add more comments and discussions here than have you do different PR now, which is more time consumed for you and probably more confusion for me and Ben later on. 

Lets make a more detailed log here and move forward.

tizianoGuadagnino:
FYI we will give the final answer about the new `AdaptiveThreshold` performances this morning, but of course testing just on public datasets might not be enough.

tizianoGuadagnino:
PS: we discuss with @benemer  and ATE change goes in fact in a different PR because we are still in the process of getting it right. Ben is working on splitting this, after that we can merge

benemer:
New PR with the ATE change is now at https://github.com/PRBonn/kiss-icp/pull/324

I reverted it here :+1:

tizianoGuadagnino (approved):
Confirmed that the results do not change significantly in any of the datasets we tested with @benemer. We can proceed to merge this.

benemer (approved):

PR #320, opened 2024-04-10
Reduce compute in RemovePointsFarFromLocation

QuentinTorg:
Remove unnecessary computation in each loop of the VoxelHashMap RemovePointsFarFromLocation function

In some situations the compiler may do this optimization, but I figured this is a pretty low risk change


nachovizzo (approved):
Fair enough, good catch. Thanks!

benemer (approved):

tizianoGuadagnino (approved):

PR #321, opened 2024-04-11
Get absolute path first in case . is data_dir

benemer:
@louis-wiesmann caught a small bug when running the `kiss_icp_pipeline` directly in the data folder. The generic data loader will use the "." as the sequence_dir, and the poses will be saved to `._poses.txt` and will therefore be hidden :laughing: 

This should fix it.


nachovizzo (approved):
Good catch kartoffelone!

tizianoGuadagnino (approved):
<3

PR #323, opened 2024-04-11
Add evaluation script 

benemer:
This adds a small script that takes two poses `.txt` files and outputs our odometry metrics.

### DIFF
--- a/cpp/kiss_icp/metrics/Metrics.cpp
+++ b/cpp/kiss_icp/metrics/Metrics.cpp
@@ -172,14 +172,17 @@ std::tuple<float, float> AbsoluteTrajectoryError(const std::vector<Eigen::Matrix
     for (size_t j = 0; j < num_poses; ++j) {
         // Apply alignement matrix
         const Eigen::Matrix4d T_estimate = T_align_trajectories * poses_result[j];
-        const Eigen::Matrix4d T_ground_truth = poses_gt[j];
-        const auto delta_T = T_estimate.inverse() * T_ground_truth;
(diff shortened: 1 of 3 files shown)


benemer:
I'm not sure about the naming, though, because this only takes the `.txt` files in KITTI format. What do you guys think?

tizianoGuadagnino:
In terms of the input `.txt` files, it is clear from the name of the script that this is isolated to the kitti metric. Anyway, I don't like the idea of the script at this stage of the project, as I think the info into the `results.log` are usually sufficient for the python users.

benemer:
Reverted [1bcd4d3](https://github.com/PRBonn/kiss-icp/pull/323/commits/1bcd4d36c9eea6282c87df29d7a8f5b27efb3358) to keep PRs more separated. This will be handled in https://github.com/PRBonn/kiss-icp/pull/324.

benemer:
> In terms of the input `.txt` files, it is clear from the name of the script that this is isolated to the kitti metric. Anyway, I don't like the idea of the script at this stage of the project, as I think the info into the `results.log` are usually sufficient for the python users.

I can also make it more general and usable for any type of pose file (kitti `.txt`, tum `txt`, `npy'). But we should first discuss whether the script should be part of the project :+1: 

It is useful if you run experiments using the Python API and want to compare the KITTI metrics across pose files from, for example, other odometry baselines. But I agree that, at least for the ATE, we can also use external packages like `evo_ape`. @nachovizzo, what is your opinion on this?

benemer:
Fine for me 👍

PR #324, opened 2024-04-12
Fix ATE

benemer:
The ATE computation was wrong, now fixed


tizianoGuadagnino (approved):
Finally this is correct

benemer:
@nachovizzo can this be merged?

We double checked the ATE values with `evo_ape` and they match now.

nachovizzo (approved):

PR #326, opened 2024-04-16
Consistent type for GT poses

benemer:
This is a minor thing, but the type annotation for the GT poses was a list of `np.array`. However, we usually return a single `np.array`.

I changed the type annotation and made the return type consistent for the Python dataloaders.


nachovizzo (approved):
Ho yeah 🕺

tizianoGuadagnino (approved):

PR #327, opened 2024-04-16
Change target names

tizianoGuadagnino:
In the current setting, if someone will `FetchContent` KISS, there might be conflicts based on target names, as `core` and `pipeline` are very common. There is probably another workaround to this, but it makes more sense to remove the aliasing and call every target `kiss_icp_<something>`. This is the safest option that should not fire back later on, I attach here an example of what I am talking about.  

image

### DIFF
--- a/cpp/kiss_icp/3rdparty/eigen/eigen.cmake
+++ b/cpp/kiss_icp/3rdparty/eigen/eigen.cmake
@@ -25,16 +25,18 @@
 # release a new version that is not 3.4. That version does not include this necessary changes:
 # - https://gitlab.com/libeigen/eigen/-/merge_requests/893/diffs
 
-set(EIGEN_BUILD_DOC OFF CACHE BOOL "Don't build Eigen docs")
-set(EIGEN_BUILD_TESTING OFF CACHE BOOL "Don't build Eigen tests")
-set(EIGEN_BUILD_PKGCONFIG OFF CACHE BOOL "Don't build Eigen pkg-config")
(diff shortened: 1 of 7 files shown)


tizianoGuadagnino:
#328 introduce the change in the build system that will also fix FetchContent

benemer (approved):

nachovizzo:
I knew this might happen sooner or later. I think the underlying problem is not the name of the targets but the lack of an installable C++ library. 

I postponed this always, but I'd say we could now fix this the proper way. Using modern cmake, and relying on exported targets instead of manually adding directories in the current build space

A probably outdated example:
https://github.com/PRBonn/vdbfusion/blob/main/src/vdbfusion/vdbfusion/cmake/Install.cmake

 In summary,  I think we should make the  C++ library installable, like is the case with all the dependencies we use

tizianoGuadagnino:
Absolutely agree with @nachovizzo, that should be the proper way. Shall we merge this and open a new branch for the _install_ ?

nachovizzo:
> Absolutely agree with @nachovizzo, that should be the proper way. Shall we merge this and open a new branch for the _install_ ?

I'm fine with it as long as we commit to do it and review it relatively soon

I will approve this so I don't block you but please let's revisit this soon. I won't have time in the next weeks to work on this particular build feature

nachovizzo (approved):

tizianoGuadagnino:
You got my word.

PR #328, opened 2024-04-16
EXCLUDE_FROM_ALL goes into the FetchContent_Declare

tizianoGuadagnino:
This fixes the `FetchContent` population error that we introduced many months before. We were checking for a `depname_Populated` without using `FetchContent_GetProperties(depname)'. This resulted in cmake error as it was trying to populate the same dependency multiple times if KISS was "_fetch contented_" by another cmake project that was also "_fetch conteting_" one of the KISS dependencies.

### DIFF
--- a/cpp/kiss_icp/3rdparty/eigen/eigen.cmake
+++ b/cpp/kiss_icp/3rdparty/eigen/eigen.cmake
@@ -32,16 +32,6 @@ set(EIGEN_BUILD_BLAS OFF CACHE BOOL "Don't build blas module")
 set(EIGEN_BUILD_LAPACK OFF CACHE BOOL "Don't build lapack module")
 
 include(FetchContent)
-FetchContent_Declare(eigen URL https://github.com/nachovizzo/eigen/archive/refs/tags/3.4.90.tar.gz)
-if(NOT eigen_POPULATED)
-  FetchContent_Populate(eigen)
-  if(${CMAKE_VERSION} GREATER_EQUAL 3.25)
-    add_subdirectory(${eigen_SOURCE_DIR} ${eigen_BINARY_DIR} SYSTEM EXCLUDE_FROM_ALL)
-  else()
-    # Emulate the SYSTEM flag introduced in CMake 3.25. Withouth this flag the compiler will
-    # consider this 3rdparty headers as source code and fail due the -Werror flag.
-    add_subdirectory(${eigen_SOURCE_DIR} ${eigen_BINARY_DIR} EXCLUDE_FROM_ALL)
-    get_target_property(eigen_include_dirs eigen INTERFACE_INCLUDE_DIRECTORIES)
-    set_target_properties(eigen PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${eigen_include_dirs}")
-  endif()
-endif()
+FetchContent_Declare(eigen URL https://github.com/nachovizzo/eigen/archive/refs/tags/3.4.90.tar.gz EXCLUDE_FROM_ALL
+                               TRUE)
+FetchContent_MakeAvailable(eigen)
--- a/cpp/kiss_icp/3rdparty/tbb/tbb.cmake
+++ b/cpp/kiss_icp/3rdparty/tbb/tbb.cmake
@@ -28,16 +28,6 @@ option(TBB_STRICT OFF)
 option(TBB_TEST OFF)
 
 include(FetchContent)
-FetchContent_Declare(tbb URL https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.8.0.tar.gz)
-if(NOT tbb_POPULATED)
-  FetchContent_Populate(tbb)
-  if(${CMAKE_VERSION} GREATER_EQUAL 3.25)
-    add_subdirectory(${tbb_SOURCE_DIR} ${tbb_BINARY_DIR} SYSTEM EXCLUDE_FROM_ALL)
-  else()
-    # Emulate the SYSTEM flag introduced in CMake 3.25. Withouth this flag the compiler will
-    # consider this 3rdparty headers as source code and fail due the -Werror flag.
-    add_subdirectory(${tbb_SOURCE_DIR} ${tbb_BINARY_DIR} EXCLUDE_FROM_ALL)
-    get_target_property(tbb_include_dirs tbb INTERFACE_INCLUDE_DIRECTORIES)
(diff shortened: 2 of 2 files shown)


tizianoGuadagnino:
Proposal to @nachovizzo: we can merge the three PRs (This one, #327, and  #benedikt/remove_personal_repos) about the build system in this one and have all the related changes in one place.

nachovizzo:
Guys I'm a bit lost with these 3 PR's

 In any case, `EXCLUDE_FROM_ALL` in the Fetchcontent was only introduced in CMake 3.28, which literally renders all these PRs wrong AFIK

tizianoGuadagnino:
I give you an order of creation:

1) Rename targets, which is where we started #327.
2) This triggers thoughts about some dependencies being populated by multiple projects (if you build KISS and [MapClosures](https://github.com/PRBonn/MapClosures) for example), which in our current state implies that the build system core dumps. We should somehow handle the case in which `Eigen` as been already populated which, with `FetchContent_MakeAvailable` does already.
3) We want to make the dependencies cmake scripts cleaner (with no personal repos) just because #329 

Thats the situation.

nachovizzo:
Thanks @tizianoGuadagnino. But were you guys aware of the cmake increase on the version???

tizianoGuadagnino:
Yes, I just wanted to be sure that it is fixable in this way and discuss it. Increasing the CMake version is a mess, and I know, but we need to find a solution to this; otherwise, people might have trouble building on top of our stuff.

There is a very ugly if else logic that can replace the populate checks that `MakeAvailable` so that we don't need `cmake` > 3.22. I just though it is worth a discussion here

nachovizzo:
> Yes, I just wanted to be sure that it is fixable in this way and discuss it. Increasing the CMake version is a mess, and I know, but we need to find a solution to this; otherwise, people might have trouble building on top of our stuff.
> 
> There is a very ugly if else logic that can replace the populate checks that `MakeAvailable` so that we don't need `cmake` > 3.22. I just though it is worth a discussion here

Ok agree, fixing the problem is fine, but I'm not in favor of bumping the version at all. I understand that this will beautify the build system, and I tried this in the past as well and always dream about that. But we should in best cases expect most users will have cmake 3.22.

One option would be to add this nice addition with 3.28 and guard it with an if/else, similar to how we do now

tizianoGuadagnino:
Mmm, I didn't think about this one. Let me have a look, then ;)

nachovizzo:
So, let me wrap up this tsunami: 

The only neccessary change is https://github.com/PRBonn/kiss-icp/pull/327/, then #328 and #329 are just proposals on "nice to have" . Given the  versioning I'd say this PR should be closed 

I'm trying to understand what really needs fixing ;)

tizianoGuadagnino:
I will say #328 is not just nice to have but necessary if someone needs to use KISS as a library.

At times, @benemer and I have received 5 PRs as _tsunami_, so I think that is fine. However, we should not close this without all agreeing that those are marginal changes (which we don't think they are). 

Fix for the versioning in this PR with the guarding that we just discuss it is coming anyways.

nachovizzo:
> I will say #328 is not just nice to have but necessary if someone needs to use KISS as a library.



> I will say #328 is not just nice to have but necessary if someone needs to use KISS as a library.
> 
> At times, @benemer and I have received 5 PRs as _tsunami_, so I think that is fine. However, we should not close this without all agreeing that those are marginal changes (which we don't think they are).
> 
> Fix for the versioning in this PR with the guarding that we just discuss it is coming anyways.

Ok I missed this:

> We were checking for a depname_Populated without using `FetchContent_GetProperties(depname)'.

I got confused sorry.

But why don't just fixing that  ?

cpp/kiss_icp/3rdparty/eigen/eigen.cmake
```diff
@@ -32,16 +32,23 @@ set(EIGEN_BUILD_BLAS OFF CACHE BOOL "Don't build blas module")
...
-    add_subdirectory(${eigen_SOURCE_DIR} ${eigen_BINARY_DIR} EXCLUDE_FROM_ALL)
-    get_target_property(eigen_include_dirs eigen INTERFACE_INCLUDE_DIRECTORIES)
-    set_target_properties(eigen PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${eigen_include_dirs}")
+if(${CMAKE_VERSION} GREATER_EQUAL 3.28)
+  FetchContent_Declare(eigen URL https://github.com/nachovizzo/eigen/archive/refs/tags/3.4.90.tar.gz EXCLUDE_FROM_ALL
+                                 SYSTEM)
+  FetchContent_MakeAvailable(eigen)
+else()
+  FetchContent_Declare(eigen URL https://github.com/nachovizzo/eigen/archive/refs/tags/3.4.90.tar.gz)
+  FetchContent_GetProperties(eigen)
```
nachovizzo, line 41: This is the real fix right? this should fix the issue without adding the extra 3.28., Or did I got it wrong?

tizianoGuadagnino: No, no, that will fix it as well; when you mention the guarding, I thought you wanted to keep the 3.28 version of things for future reference. I also misunderstood then, next time, phone call.

nachovizzo: Maybe worth leaving a comment for the future with a link to this PR, as in a few years maybe cmake 3.28 is mainstream and we can replace all these madness with those 2 lines

nachovizzo (approved):
amashing

cpp/kiss_icp/3rdparty/eigen/eigen.cmake
```diff
@@ -33,6 +33,7 @@ set(EIGEN_BUILD_LAPACK OFF CACHE BOOL "Don't build lapack module")
 
 include(FetchContent)
 FetchContent_Declare(eigen URL https://github.com/nachovizzo/eigen/archive/refs/tags/3.4.90.tar.gz)
+FetchContent_GetProperties(eigen)
 if(NOT eigen_POPULATED)
```
nachovizzo, line 37: This basically means this was literally `if (TRUE)` since the beginning of time

tizianoGuadagnino: Yeees, when I found out my brain just got bananas

tizianoGuadagnino: But it is now done

benemer (approved):
Awesome!

:surfer: :surfer: :surfer:

PR #332, opened 2024-04-18
Add empty pipeline constructor

benemer:
This PR allows you to initialize the pipeline without a config at hand by using the default KISSConfig. In that case, you still want to access the internal config, so we also added an accessor to it.

### DIFF
--- a/cpp/kiss_icp/core/Registration.hpp
+++ b/cpp/kiss_icp/core/Registration.hpp
@@ -22,6 +22,8 @@
 // SOFTWARE.
 #pragma once
 
+#include <oneapi/tbb/global_control.h>
+
 #include <Eigen/Core>
 #include <sophus/se3.hpp>
 #include <vector>
@@ -41,6 +43,8 @@ struct Registration {
 
     int max_num_iterations_;
     double convergence_criterion_;
-    int max_num_threads_;
+    // This attribute requires static to be able to manipulate the max
+    // concurrency from TBB across multiple instances of this class
+    static tbb::global_control tbb_control_settings;
 };
 }  // namespace kiss_icp
--- a/cpp/kiss_icp/core/Registration.hpp
+++ b/cpp/kiss_icp/core/Registration.hpp
@@ -22,7 +22,8 @@
 // SOFTWARE.
 #pragma once
 
-#include <oneapi/tbb/global_control.h>
+#include <tbb/global_control.h>
+#include <tbb/info.h>
 
 #include <Eigen/Core>
 #include <sophus/se3.hpp>
--- a/cpp/kiss_icp/core/Registration.hpp
+++ b/cpp/kiss_icp/core/Registration.hpp
@@ -44,8 +44,8 @@ struct Registration {
 
     int max_num_iterations_;
     double convergence_criterion_;
-    // This attribute requires static to be able to manipulate the max
-    // concurrency from TBB across multiple instances of this class
-    static tbb::global_control tbb_control_settings;
+    // This attribute is needed to be able to manipulate the max
+    // concurrency from TBB this class
+    tbb::global_control tbb_control_settings;
 };
 }  // namespace kiss_icp
--- a/cpp/kiss_icp/pipeline/KissICP.hpp
+++ b/cpp/kiss_icp/pipeline/KissICP.hpp
@@ -58,6 +58,7 @@ public:
     using Vector3dVectorTuple = std::tuple<Vector3dVector, Vector3dVector>;
 
 public:
+    explicit KissICP() : KissICP(KISSConfig()) {}
     explicit KissICP(const KISSConfig &config)
         : config_(config),
           registration_(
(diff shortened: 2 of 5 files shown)


tizianoGuadagnino (dismissed):

cpp/kiss_icp/pipeline/KissICP.hpp
```diff
@@ -58,6 +58,7 @@ class KissICP {
     using Vector3dVectorTuple = std::tuple<Vector3dVector, Vector3dVector>;
 
 public:
+    explicit KissICP() : KissICP(KISSConfig()) {}
```
nachovizzo, line 61: This was intentionally deleted in https://github.com/PRBonn/kiss-icp/pull/294

cpp/kiss_icp/pipeline/KissICP.hpp
```diff
@@ -74,6 +75,7 @@ class KissICP {
     std::vector<Eigen::Vector3d> LocalMap() const { return local_map_.Pointcloud(); };
     Sophus::SE3d pose() const { return last_pose_; }
     Sophus::SE3d delta() const { return last_delta_; }
+    const KISSConfig &config() const { return config_; }
```
nachovizzo, line 78: This is fine :+1:

cpp/kiss_icp/pipeline/KissICP.hpp
benemer: Okay, then I didn't fully understand what the problem in https://github.com/PRBonn/kiss-icp/pull/294 was. Can you explain again? Is this about the tbb max number of threads being set to default at compile time if a default constructor is provided?

cpp/kiss_icp/pipeline/KissICP.hpp
benemer: Okay, but without the empty constructed it is not needed, because then you need to provide a config anyway to initialize a pipeline :(

cpp/kiss_icp/pipeline/KissICP.hpp
nachovizzo: Basically yes, would be a matter of taste.  I kind of liked the idea of not allowing to construction the pipeline with default parameters.

Not a big deal though, but then we shall remove the hack off the TBB number of threads and make it a registration member, this uglyfy quite a bit the API while being more correct at the same time.

cpp/kiss_icp/core/Registration.hpp
```diff
@@ -41,6 +43,8 @@ struct Registration {
 
     int max_num_iterations_;
     double convergence_criterion_;
-    int max_num_threads_;
+    // This attribute requires static to be able to manipulate the max
+    // concurrency from TBB across multiple instances of this class
+    static tbb::global_control tbb_control_settings;
```
tizianoGuadagnino, line 48: @nachovizzo do you mean something like this?

tizianoGuadagnino: I don't think the default constructor is really needed from our side, I think I wanted mostly to make it right, although a bit uglier. I think we can vote and decide altogether because this is not critical. But let's make it correct more so that beautiful but hidden.

nachovizzo: All good, then if this is the target of the PR we should rename the PR or close/reopen. Otherwise, It's extremely confusing for me that open this at different intervals.

I'm happy if there is quorum to remove the static hack, but let's vote

benemer: > We can't know the max range of your sensor at compile time, so we "force" you to specify this dynamically.

Well, this we don't really enforce because we have default configs in both the pipeline header and the ROS launch file. The main idea of this PR was more: If you provide a default config, you can also provide a default constructor.

If we agree not to allow a default constructor, we should be more consistent and also not provide a default config. Otherwise, people can anyway just initialize with the default config.

nachovizzo: You are right, we are not enforcing per-se. In fact, I don't believe anyone actually realizes that we have some parameters to tune 🤔 which is a success I'd say.

The main difference for the C++ side is that it's more likely users will indeed instantiate an object of the pipeline class, in contrast to the python case where we don't really expect that use case and people will use it more as a black-box system. In fact, one can't bake a `KissICP` object in python without providing a config
image

If I'm writing C++ code, I'd love to see this error at compile time, and at least be conisuous of constructing the object with a default config, which I don't even want to look into maybe, but understand that there is a config.
In either case, it's a matter of taste and decision

benemer: Okay, that's a fair point.

Regarding making the tbb control setting an attribute I don't have a strong opinion, so I am fine with either renaming this PR or closing it.

tizianoGuadagnino: It seems none of us strongly need/want/desire the tbb control as an attribute or the default constructor. Let's close this PR for now and reevaluate later if something changes. thanks, guys

PR #338, opened 2024-05-06
HeLiPR: Make sure chunk size is sufficient

benemer:
When reading HeLiPR data, in some sequences (for example `Roundabout02`) the last chunk of data is incomplete and Kiss fails with

```shell
error: unpack_from requires a buffer of at least 999400 bytes for unpacking 25 bytes at offset 999375 (actual buffer size is 999383)
```

This PR fixes this by ensuring we have enough data to read in the last chunk.


nachovizzo:
Should I download this dataset 😂?

nachovizzo (approved):

tizianoGuadagnino (approved):

Issue #339, opened 2024-05-09
Inconsistent unit (rad vs deg) in c++ and python


wing-kit:
https://github.com/PRBonn/kiss-icp/blob/main/cpp/kiss_icp/metrics/Metrics.cpp#L152
In the implementation part it is in rad/m.

https://github.com/PRBonn/kiss-icp/blob/main/python/kiss_icp/metrics.py#L30

In the rendering it says it is in deg/m.
https://github.com/PRBonn/kiss-icp/blob/main/python/kiss_icp/pipeline.py#L175

benemer:
See https://github.com/PRBonn/kiss-icp/pull/340

PR #340, opened 2024-05-09
from deg/m to rad/m

wing-kit:
https://github.com/PRBonn/kiss-icp/issues/339


benemer:
No, in https://github.com/PRBonn/kiss-icp/blob/4225185d1c4a2d48c5f4f47063422b2cd35802cd/cpp/kiss_icp/metrics/Metrics.cpp#L152 we divide by pi and multiply with 180 degrees, so deg/m should be correct. Or am I missing anything?

benemer:
I will close this for now, but feel free to comment and re-open it if needed!

Issue #343, opened 2024-05-17
ZeroDivisionError in pipeline.py in  ‎OdometryPipeline._get_fps function


karthikbolla:
![image](https://github.com/PRBonn/kiss-icp/assets/110156560/4027af0d-a8fe-498f-a69b-145098a252ad)

In the above function, if total_time_s becomes zero due to dataset issues, it throws a ZeroDivisionError: division by zero.

PR #344, opened 2024-05-17
Handled Zero Division Error in pipeline.py

karthikbolla:
This pull request addresses a  division by zero error in the `_get_fps()` function. I've implemented a check to handle the case where `total_time_s` is zero to prevent the error. Additionally, I've modified the calculation of `avg_ms` to ensure it doesn't result in a division by zero. 

Changes Made:
- Added a conditional check in `_get_fps()` to handle division by zero.
- Adjusted the calculation of `avg_ms` to prevent division by zero.

Fixes: #343

### DIFF
--- a/python/kiss_icp/pipeline.py
+++ b/python/kiss_icp/pipeline.py
@@ -179,12 +179,21 @@ class OdometryPipeline:
         # Run timing metrics evaluation, always
         def _get_fps():
             total_time_s = sum(self.times) * 1e-9
-            return float(len(self.times) / total_time_s)
+            if total_time_s == 0:
+                return 0  
+            else:
+                return float(len(self.times) / total_time_s)
 
-        avg_fps = int(np.ceil(_get_fps()))
-        avg_ms = int(np.ceil(1e3 * (1 / _get_fps())))
+        fps = _get_fps()
+        if fps == 0:
+            avg_fps = 0
+            avg_ms = 0
+        else:
+            avg_fps = int(np.ceil(fps))
+            avg_ms = int(np.ceil(1e3 / fps))
         self.results.append(desc="Average Frequency", units="Hz", value=avg_fps, trunc=True)
         self.results.append(desc="Average Runtime", units="ms", value=avg_ms, trunc=True)
+
 
     def _write_log(self):
         if not self.results.empty():
--- a/python/kiss_icp/pipeline.py
+++ b/python/kiss_icp/pipeline.py
@@ -179,18 +179,11 @@ class OdometryPipeline:
         # Run timing metrics evaluation, always
         def _get_fps():
             total_time_s = sum(self.times) * 1e-9
-            if total_time_s == 0:
-                return 0  
-            else:
-                return float(len(self.times) / total_time_s)
+            return float(len(self.times) / total_time_s) if total_time_s > 0 else 0
 
         fps = _get_fps()
-        if fps == 0:
-            avg_fps = 0
-            avg_ms = 0
-        else:
-            avg_fps = int(np.ceil(fps))
-            avg_ms = int(np.ceil(1e3 / fps))
+        avg_fps = int(np.floor(fps))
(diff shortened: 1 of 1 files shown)


python/kiss_icp/pipeline.py
```diff
@@ -179,13 +179,22 @@ def _run_evaluation(self):
...
         def _get_fps():
             total_time_s = sum(self.times) * 1e-9
-            return float(len(self.times) / total_time_s)
-
-        avg_fps = int(np.ceil(_get_fps()))
-        avg_ms = int(np.ceil(1e3 * (1 / _get_fps())))
+            if total_time_s == 0:
+                return 0  
+            else:
+                return float(len(self.times) / total_time_s)
```
benemer, line 185:
```suggestion
            return float(len(self.times) / total_time_s) if total_time_s > 0 else 0
```

python/kiss_icp/pipeline.py
```diff
@@ -179,13 +179,22 @@ def _run_evaluation(self):
...
+
+        fps = _get_fps()
+        if fps == 0:
+            avg_fps = 0
+            avg_ms = 0
+        else:
+            avg_fps = int(np.ceil(fps))
+            avg_ms = int(np.ceil(1e3 / fps))
```
benemer, line 193:
```suggestion
        avg_fps = int(np.floor(fps))
        avg_ms = int(np.ceil(1e3 / fps)) if fps > 0 else 0
```

This again saves some lines and reports a more conservative fps and runtime.

benemer:
Thanks for the fix! Two cosmetic suggestions from my side, just to keep things compact.

nachovizzo:
Thanks for the fix, it is definetly helpful.

The bigger question here would be, though, how you end up with a 0.000 runtime. You might have a problem somewhere else. I agree that is not the right place to explode, but something else went wrong here

benemer:
Is this PR still active? We should discuss which runtime should be reported if the total time is zero. I have two proposals:

1. Make it NaN, this also occurs sometimes for the relative error
2. Do not show it in the final logs

What do you think?

karthikbolla:
> Is this PR still active? We should discuss which runtime should be reported if the total time is zero. I have two proposals:
> 
> 1. Make it NaN, this also occurs sometimes for the relative error
> 2. Do not show it in the final logs
> 
> What do you think?

Is there any other way than returning 0?

benemer:
How about

```diff
diff --git a/python/kiss_icp/pipeline.py b/python/kiss_icp/pipeline.py
index 52cb7a7..46b9d4d 100644
--- a/python/kiss_icp/pipeline.py
+++ b/python/kiss_icp/pipeline.py
@@ -184,9 +184,10 @@ class OdometryPipeline:
         fps = _get_fps()
         avg_fps = int(np.floor(fps))
         avg_ms = int(np.ceil(1e3 / fps)) if fps > 0 else 0
-        self.results.append(desc="Average Frequency", units="Hz", value=avg_fps, trunc=True)
-        self.results.append(desc="Average Runtime", units="ms", value=avg_ms, trunc=True)
 
+        if avg_fps > 0:
+            self.results.append(desc="Average Frequency", units="Hz", value=avg_fps, trunc=True)
+            self.results.append(desc="Average Runtime", units="ms", value=avg_ms, trunc=True)
 
     def _write_log(self):
         if not self.results.empty():
```

Wit this, we don't have think about what the `correct` return value would be.

nachovizzo:
@karthikbolla would you please do as @benemer suggested? you can also give us access to commit there so we also ack your contribution


thanks

karthikbolla:
@nachovizzo @benemer  I've raised a new commit based on @benemer's  suggestions and also granted access to the fork. 

Thank you.

benemer (dismissed):
Great, thank you for your contribution!

tizianoGuadagnino (dismissed):

nachovizzo (approved):

PR #345, opened 2024-06-04
Fix kiss_icp_dump_config executable

benemer:
The function called with the `kiss_icp_dump_config` executable does not exist, I changed it to the correct `write_config` function.


nachovizzo (approved):
Do we use this function? 🤔 I forgot we had it there

benemer:
The `write_dump_config` does not exist; the `write_config` is used by the pipeline to save the config after the run. The executable we added also uses this function to dump a default config as a template if you want to modify some parameters.

tizianoGuadagnino (approved):

PR #349, opened 2024-06-24
Some memory saving is possible when getting points from voxel hash map

l00p3:
In these two methods of VoxelHasMap the memory for vectors "points" is reserved but never cleared if not completely filled. I checked and size and capacity often doesn't match, resulting in some small waste of memory.
Also, in the method PointCloud() std::for_each is probably better and emplace_back instead of push_back, just to be pedantic. The static cast for max_points_per_voxel_ from int to size_t is there just to be consistent with the method GetPoints().

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -24,7 +24,6 @@
 
 #include <Eigen/Core>
 #include <algorithm>
-#include <limits>
 #include <tuple>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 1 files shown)


Pyrestone:
Please make sure this change doesn't negatively affect runtime performance. Constantly re-allocating that chunk of memory to different sizes might be very detrimental, and a poor exchange for some intermediate, temporary memory savings.

l00p3:
Yes, I am currently running some experiment to check it. Thanks.

l00p3:
> Please make sure this change doesn't negatively affect runtime performance. Constantly re-allocating that chunk of memory to different sizes might be very detrimental, and a poor exchange for some intermediate, temporary memory savings.

These the results on a couple of sequences (should be enough). No change on my machine.
merged

Theoretically makes sense, shrink_to_fit() is just a request to change the capacity. The compiler optimizes if it is actually needed or not. Also, in case of capacity change the complexity is linear in the dimension of the vector because it just goes at the end of the vector and changes the "opinter" to the last "available element", no copy should happen.

tizianoGuadagnino (approved):

benemer (approved):

PR #353, opened 2024-07-02
mcap reader: Use extracted topic

benemer:
If no topic is provided, the `self.check_topic` is supposed to extract it, for example, if there is only one message of type `PointCloud2`. However, we did not pass the inferred topic. This is the fix.

### DIFF
--- a/python/kiss_icp/datasets/mcap.py
+++ b/python/kiss_icp/datasets/mcap.py
@@ -46,7 +46,7 @@ class McapDataloader:
         self.summary = self.bag.get_summary()
         self.topic = self.check_topic(topic)
         self.n_scans = self._get_n_scans()
-        self.msgs = read_ros2_messages(mcap_file, topics=topic)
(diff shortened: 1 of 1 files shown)


nachovizzo (approved):
How come 🤔 I thought it was working but maybe I'm confused with lidar visualizer

benemer:
Same thing there actually: https://github.com/PRBonn/lidar-visualizer/blob/3884578c3943c28b4bfba4ad8ca96a7fb3eca27a/src/lidar_visualizer/datasets/mcap.py#L51C1-L52C1

How to reproduce the error I faced: I had a `mcap` file with just one PointCloud2 message and did not specify the topic. Doing this, there was no warning that the topic should be passed (because it could extract it), but the resulting topic passed to the reader was empty.

So probably this did not fail before because we always passed the topic.

saurabh1002 (approved):

PR #355, opened 2024-07-02
Add timestamps to mcap

benemer:
To correctly save the poses to the `tum` format, we need the timestamps of the scan. For the `rosbag` reader we extract these from the message header. This PR is implementing the same for the `mcap` reader.

Depends on
https://github.com/PRBonn/kiss-icp/pull/353


nachovizzo (dismissed):
Perfect, thanks Benedetto

saurabh1002 (dismissed):
Looks good to me

benemer:
Merged after approval of @nachovizzo and @saurabh1002 :)

PR #358, opened 2024-07-04
Remove redundant Modulus operation for Voxel Hash function

saurabh1002:
1. Remove the modulus operation `(1 << 20) - 1 & <hash>` from the custom Hashing operator.
2. Initialize the robin map with a higher bucket size, change the default load factors.

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.hpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.hpp
@@ -46,14 +46,18 @@ struct VoxelHashMap {
     struct VoxelHash {
         size_t operator()(const Voxel &voxel) const {
             const uint32_t *vec = reinterpret_cast<const uint32_t *>(voxel.data());
-            return ((1 << 20) - 1) & (vec[0] * 73856093 ^ vec[1] * 19349669 ^ vec[2] * 83492791);
+            return (vec[0] * 73856093 ^ vec[1] * 19349669 ^ vec[2] * 83492791);
         }
     };
 
     explicit VoxelHashMap(double voxel_size, double max_distance, int max_points_per_voxel)
         : voxel_size_(voxel_size),
           max_distance_(max_distance),
-          max_points_per_voxel_(max_points_per_voxel) {}
+          max_points_per_voxel_(max_points_per_voxel),
+          map_(1 << 16) {
+        map_.min_load_factor(0.05f);
+        map_.max_load_factor(0.75f);
+    }
 
     inline void Clear() { map_.clear(); }
     inline bool Empty() const { return map_.empty(); }


benemer:
Can you give some pointers to the documentation where this can be found? Especially that the modulus operation is done on the hash map side

saurabh1002:
#### The Hashing function

The modulus operation in the hashing operator of the custom hash function is actually not required by the Robin Map implementation. They perform that operation internally based on the current bucket size, which keeps on changing based on current load_factor of the hash table. See the figure below where I backtrace to that modulus operation in Robin Map library.

In fact, the current bucket size that we have, through the hash operator, is 2^20 buckets. This can cause an issue, when internally the robin map grows to more than 2^20 buckets, in which case our hash function will limit the robin map bucket computation. Otherwise, when the internal bucket size is below 2^20, our additional modulus operation is just redundant.

image

saurabh1002:
#### Default load factor params
The robin map internally resizes the hash table based on the load_factor thresholds.
1. It upscales the bucket size by a factor of **2** when load_factor goes above **0.5**, which is an expensive operation as it recomputes all the hashes and the buckets they are assigned to.
2. It downscales the bucket size by a factor of **2** again, when the load_factor goes below **0.0**, which means basically never.

It is better if we already initialize the robin map with a sufficient bucket size, along with increasing the maximum load factor threshold for upscaling to **0.75**. This will avoid frequent rehashing of the map, especially at the beginning of any sequence.
Also, we can set the minimum load factor threshold for downscaling to **0.05** to have some sort of downscaling when operating in narrow regions with lower FOV, and avoid costly iterations through the map, for example when querying for nearest neighbors, or generating pointcloud from local map.

saurabh1002:
image

saurabh1002:
#### Test on MulRAN Sequences
image

saurabh1002:
#### Apollo Dataset
image

cpp/kiss_icp/core/VoxelHashMap.hpp
```diff
@@ -46,14 +46,18 @@ struct VoxelHashMap {
...
-            return ((1 << 20) - 1) & (vec[0] * 73856093 ^ vec[1] * 19349669 ^ vec[2] * 83492791);
+            return (vec[0] * 73856093 ^ vec[1] * 19349669 ^ vec[2] * 83492791);
         }
     };
 
     explicit VoxelHashMap(double voxel_size, double max_distance, int max_points_per_voxel)
         : voxel_size_(voxel_size),
           max_distance_(max_distance),
-          max_points_per_voxel_(max_points_per_voxel) {}
+          max_points_per_voxel_(max_points_per_voxel),
```
nachovizzo, line 56: Is it too much to ask to split this into a separate PR? This is to avoid chasing ghosts in the future. If we merge this it will appear in the commit history as ``Remove redundant Modulus operation for Voxel Hash function

I can also do it if you give me access to your fork

tizianoGuadagnino: I just did it, some parts where kind of magic numbers there so I decide to keep just the explainable ones in the new PR.

tizianoGuadagnino (approved):

nachovizzo (approved):
Fantastic @saurabh1002 thanks !

PR #359, opened 2024-07-09
Initialize hash buckets

tizianoGuadagnino:
This split  #358 into two parts, this one regards the initial bucket size for the ```robin_map```.

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.hpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.hpp
@@ -53,7 +53,9 @@ struct VoxelHashMap {
     explicit VoxelHashMap(double voxel_size, double max_distance, int max_points_per_voxel)
         : voxel_size_(voxel_size),
           max_distance_(max_distance),
-          max_points_per_voxel_(max_points_per_voxel) {}
+          max_points_per_voxel_(max_points_per_voxel),
+          map_(1 << 20) {}  // <- Number of initial buckets into the robin map, to avoid rehashing
+                            // (a.k.a big copy operation) in the default constructor this is zero.
 
     inline void Clear() { map_.clear(); }
     inline bool Empty() const { return map_.empty(); }


nachovizzo:
Hello nacho of the future, I knew you will ask this, so here is a brief explanation:

As far as I can trace, this is happening under the hood.

```cpp
// ....
class robin_hash : private Hash, private KeyEqual, private GrowthPolicy {

  using buckets_container_type = std::vector<bucket_entry, buckets_allocator>;

 // ....
 private:
  buckets_container_type m_buckets_data;

  // ....
  robin_hash(size_type bucket_count, const Hash& hash, const KeyEqual& equal,
             const Allocator& alloc,
             float min_load_factor = DEFAULT_MIN_LOAD_FACTOR,
             float max_load_factor = DEFAULT_MAX_LOAD_FACTOR)
      : Hash(hash),
        KeyEqual(equal),
        GrowthPolicy(bucket_count),
        m_buckets_data(bucket_count, alloc),                                                   // this
        m_buckets(m_buckets_data.empty() ? static_empty_bucket_ptr() : m_buckets_data.data()), // this

        m_bucket_count(bucket_count),
        m_nb_elements(0),
        m_grow_on_next_insert(false),
        m_try_shrink_on_next_insert(false) {
    if (bucket_count > max_bucket_count()) {
      TSL_RH_THROW_OR_TERMINATE(std::length_error,
                                "The map exceeds its maximum bucket count.");
    }

    if (m_bucket_count > 0) {
      tsl_rh_assert(!m_buckets_data.empty());
      m_buckets_data.back().set_as_last_bucket();
    }

    this->min_load_factor(min_load_factor);
    this->max_load_factor(max_load_factor);
  }

};
```

This translates to the fact that when `bucket_count != 0` then, there will be some runtime allocations happening right after we construct the hash table:

```cpp
        m_buckets_data(bucket_count, alloc),
        m_buckets(m_buckets_data.empty() ? static_empty_bucket_ptr() : m_buckets_data.data()),
```

Which, under the hood, will create a vector with 1 million ( `1048576` ) default constructed elements, which, in our case, is an `std::pair<Voxel, VoxelBlock>`

```cpp
/**
 *  @brief  Creates a %vector with default constructed elements.
 *  @param  __n  The number of elements to initially create.
 *  @param  __a  An allocator.
 *
 *  This constructor fills the %vector with @a __n default
 *  constructed elements.
 */
explicit
vector(size_type __n, const allocator_type& __a = allocator_type())
: _Base(_S_check_init_len(__n, __a), __a)
{ _M_default_initialize(__n); }
```

In the end, it's like pre-allocating some buckets when we initialize the map instead of doing it on the fly. Do we want to do this? (it makes sense to me) but I'm just trying to figure out how much we gain with this :)

tizianoGuadagnino:
At the beginning of the local map construction, this will change quite a bit because we will not rehash even once (ideally). Once the buckets are filled, there is essentially no difference. To my eyes, it is just upsides at the beginning, so not much, probably, but it looks good.

tizianoGuadagnino:
We could also control when to do the rehashing in terms of the percentage of buckets filled, but it will require the introduction of the same magic numbers that I don't particularly enjoy.

nachovizzo:
After analyzing this in a very naive way (there are better ways to perform memory consumption profiling), these are the results.

## Instrumentation code

Inject a syscall to the kernel when constructing the object and at each iteration (when adding points to the  map).

I randomly picked Kitti 00 and run the full sequence,

```diff 
diff --git a/cpp/kiss_icp/core/VoxelHashMap.cpp b/cpp/kiss_icp/core/VoxelHashMap.cpp
index 65790fb..0f89d55 100644
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -84,6 +84,7 @@ void VoxelHashMap::AddPoints(const std::vector<Eigen::Vector3d> &points) {
             map_.insert({voxel, VoxelBlock{{point}, max_points_per_voxel_}});
         }
     });
+    printMemoryUsage();
 }
 
 void VoxelHashMap::RemovePointsFarFromLocation(const Eigen::Vector3d &origin) {
diff --git a/cpp/kiss_icp/core/VoxelHashMap.hpp b/cpp/kiss_icp/core/VoxelHashMap.hpp
index 202f661..a7cb58b 100644
--- a/cpp/kiss_icp/core/VoxelHashMap.hpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.hpp
@@ -26,12 +26,22 @@
 // should at least acknoowledge the work from CT-ICP by giving a star on GitHub
 #pragma once
 
+#include <sys/resource.h>
+#include <sys/time.h>
 #include <tsl/robin_map.h>
 
 #include <Eigen/Core>
+#include <iostream>
 #include <sophus/se3.hpp>
 #include <vector>
 
+inline void printMemoryUsage() {
+    struct rusage usage;
+    getrusage(RUSAGE_SELF, &usage);
+    std::cout << "Memory usage: " << static_cast<double>(usage.ru_maxrss) / 1024. << " MiB"
+              << std::endl;
+}
+
 namespace kiss_icp {
 struct VoxelHashMap {
     using Voxel = Eigen::Vector3i;
@@ -53,9 +63,9 @@ struct VoxelHashMap {
     explicit VoxelHashMap(double voxel_size, double max_distance, int max_points_per_voxel)
         : voxel_size_(voxel_size),
           max_distance_(max_distance),
+        printMemoryUsage();
+    }
 
     inline void Clear() { map_.clear(); }
     inline bool Empty() const { return map_.empty(); }

```

## main branch
image

## this branch
image

## Analyis

It allocates much more memory at the beginning, which is fine, as it's also a "side effect." You pay for the memory you will use at the very beginning. But I don't observe a real flattening after all, which makes me wonder if it really helps or not

nachovizzo:
Ran the same experiment with BPearl lidars, 5000 scans

I don't see any benefit in adding this now. Please let me know if I'm missing something or getting self-confused along  the way
## main

image

## this branch
image

tizianoGuadagnino:
I guess another thing that we might not see now is what happens when we have a smaller voxel size (```0.1``` or so), at that point, I can assume the rehashing is gonna happen more frequently in the _transient_ of the local map, so we might see a bigger difference at the beginning. I can also test this one

tizianoGuadagnino:
> I guess another thing that we might not see now is what happens when we have a smaller voxel size (`0.1` or so); at that point, I can assume the rehashing is gonna happen more frequently in the _transient_ of the local map so that we might see a bigger difference at the beginning. I can also test this one

Actually, after a 5 minutes meditation, I realize that even if that happens it is nothing critical that requires us to make this change. We cannot predict the size of the map, and so some rehashing is gonna happen and that is fine as this does not impact the performances. I will close this now and reopen if we find use for it later on down the line, thanks @saurabh1002 to let us think about this.

PR #360, opened 2024-07-09
[WIP] Downsampling through robin_set

tizianoGuadagnino:
This PR tries to maintain the classical behavior of our beloved ```VoxelDownsample```while taking what is good from  #347, a.k.a. avoiding the double ```for``` loop over the LiDAR frame and hash table. 

All the credit for this change goes to @l00p3. Once it is ready to merge (if), I will ensure you are between the contributors.

I will start the testing process today and come back to you, @nachovizzo @benemer feel free to be ```-Wpedantic```

### DIFF
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -23,7 +23,7 @@
 #include "Preprocessing.hpp"
 
 #include <tbb/parallel_for.h>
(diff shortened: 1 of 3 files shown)


tizianoGuadagnino:
The difference in performances between the two versions is nothing, and I feel the code for the random sampling + ```robin_set``` is much more at the end, and it is not worth the space. We will keep the (modernized) original version (hopefully #362 ) and think about this later.

PR #362, opened 2024-07-09
Revise Voxel functionalities in VoxelHashMap and Preprocessing

tizianoGuadagnino:
This unify the usage of ```Voxel``` and ```VoxelHash``` between ```Preprocessing``` and ```VoxelHashMap```, so that all the ```using``` are in one and one place only.

### DIFF
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -35,7 +35,7 @@
 
 namespace kiss_icp {
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
-                                             double voxel_size) {
+                                             const double voxel_size) {
     tsl::robin_map<Voxel, Eigen::Vector3d> grid;
     grid.reserve(frame.size());
     std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -32,12 +32,30 @@
 #include <sophus/se3.hpp>
 #include <vector>
 
+namespace {
+using Voxel = kiss_icp::VoxelHashMap::Voxel;
+using VoxelHash = kiss_icp::VoxelHashMap::VoxelHash;
+
+Voxel PointToVoxel(const Eigen::Vector3d &point, double voxel_size) {
+    return Voxel(static_cast<int>(std::floor(point.x() / voxel_size)),
+                 static_cast<int>(std::floor(point.y() / voxel_size)),
+                 static_cast<int>(std::floor(point.z() / voxel_size)));
+}
+}  // namespace
+
 namespace kiss_icp {
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
                                              double voxel_size) {
-    VoxelHashMap grid(voxel_size, 1.0, 1);
-    grid.AddPoints(frame);
-    return grid.Pointcloud();
+    tsl::robin_map<Voxel, Eigen::Vector3d, VoxelHash> grid;
+    grid.reserve(frame.size());
+    std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
+        const auto voxel = PointToVoxel(point, voxel_size);
+        if (!grid.contains(voxel)) grid.insert({voxel, point});
+    });
+    std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
+    std::transform(grid.begin(), grid.end(), frame_dowsampled.begin(),
+                   [](const auto &voxel_and_point) { return voxel_and_point.value(); });
+    return frame_dowsampled;
 }
 
 std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &frame,
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -28,25 +28,14 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <cmath>
-#include <kiss_icp/core/VoxelHashMap.hpp>
+#include <kiss_icp/core/VoxelUtils.hpp>
 #include <sophus/se3.hpp>
 #include <vector>
-
-namespace {
-using Voxel = kiss_icp::VoxelHashMap::Voxel;
-using VoxelHash = kiss_icp::VoxelHashMap::VoxelHash;
-
-Voxel PointToVoxel(const Eigen::Vector3d &point, double voxel_size) {
-    return Voxel(static_cast<int>(std::floor(point.x() / voxel_size)),
-                 static_cast<int>(std::floor(point.y() / voxel_size)),
-                 static_cast<int>(std::floor(point.z() / voxel_size)));
-}
-}  // namespace
 
 namespace kiss_icp {
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
                                              double voxel_size) {
-    tsl::robin_map<Voxel, Eigen::Vector3d, VoxelHash> grid;
+    tsl::robin_map<Voxel, Eigen::Vector3d> grid;
     grid.reserve(frame.size());
     std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
         const auto voxel = PointToVoxel(point, voxel_size);
@@ -54,7 +43,7 @@ std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> 
     });
     std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
     std::transform(grid.begin(), grid.end(), frame_dowsampled.begin(),
-                   [](const auto &voxel_and_point) { return voxel_and_point.value(); });
+                   [](const auto &voxel_and_point) { return voxel_and_point.second; });
     return frame_dowsampled;
 }
 
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -28,9 +28,10 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <cmath>
-#include <kiss_icp/core/VoxelUtils.hpp>
 #include <sophus/se3.hpp>
 #include <vector>
+
+#include "VoxelUtils.hpp"
 
 namespace kiss_icp {
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -43,9 +43,11 @@ std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> 
         const auto voxel = PointToVoxel(point, voxel_size);
         if (!grid.contains(voxel)) grid.insert({voxel, point});
     });
-    std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
-    std::transform(grid.cbegin(), grid.cend(), frame_dowsampled.begin(),
-                   [](const auto &voxel_and_point) { return voxel_and_point.second; });
+    std::vector<Eigen::Vector3d> frame_dowsampled;
+    frame_dowsampled.reserve(grid.size());
+    std::for_each(grid.cbegin(), grid.cend(), [&](const auto &voxel_and_point) {
+        frame_dowsampled.emplace_back(voxel_and_point.second);
+    });
     return frame_dowsampled;
 }
 
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -35,18 +35,20 @@
 
 namespace kiss_icp {
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
-                                             const double voxel_size) {
-    tsl::robin_map<Voxel, Eigen::Vector3d> grid;
+                                             double voxel_size) {
+    tsl::robin_map<Voxel, Eigen::Vector3d, VoxelHash> grid;
     grid.reserve(frame.size());
-    std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
+    for (const auto &point : frame) {
         const auto voxel = PointToVoxel(point, voxel_size);
-        if (!grid.contains(voxel)) grid.insert({voxel, point});
-    });
+        if (grid.contains(voxel)) continue;
+        grid.insert({voxel, point});
+    }
     std::vector<Eigen::Vector3d> frame_dowsampled;
     frame_dowsampled.reserve(grid.size());
-    std::for_each(grid.cbegin(), grid.cend(), [&](const auto &voxel_and_point) {
-        frame_dowsampled.emplace_back(voxel_and_point.second);
-    });
+    for (const auto &[voxel, point] : grid) {
+        (void)voxel;
+        frame_dowsampled.emplace_back(point);
+    }
     return frame_dowsampled;
 }
 
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -32,34 +32,12 @@
 #include <sophus/se3.hpp>
 #include <vector>
 
-namespace {
-using Voxel = kiss_icp::VoxelHashMap::Voxel;
-using VoxelHash = kiss_icp::VoxelHashMap::VoxelHash;
-
-Voxel PointToVoxel(const Eigen::Vector3d &point, double voxel_size) {
-    return Voxel(static_cast<int>(std::floor(point.x() / voxel_size)),
-                 static_cast<int>(std::floor(point.y() / voxel_size)),
-                 static_cast<int>(std::floor(point.z() / voxel_size)));
-}
-}  // namespace
-
 namespace kiss_icp {
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
                                              double voxel_size) {
-    tsl::robin_map<Voxel, Eigen::Vector3d, VoxelHash> grid;
-    grid.reserve(frame.size());
-    for (const auto &point : frame) {
-        const auto voxel = PointToVoxel(point, voxel_size);
-        if (grid.contains(voxel)) continue;
-        grid.insert({voxel, point});
-    }
-    std::vector<Eigen::Vector3d> frame_dowsampled;
-    frame_dowsampled.reserve(grid.size());
-    for (const auto &[voxel, point] : grid) {
-        (void)voxel;
-        frame_dowsampled.emplace_back(point);
-    }
-    return frame_dowsampled;
+    VoxelHashMap grid(voxel_size, 1.0, 1);
+    grid.AddPoints(frame);
+    return grid.Pointcloud();
 }
 
 std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &frame,
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -28,7 +28,6 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <cmath>
-#include <iterator>
 #include <sophus/se3.hpp>
 #include <vector>
 
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -28,6 +28,7 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <cmath>
+#include <iterator>
 #include <sophus/se3.hpp>
 #include <vector>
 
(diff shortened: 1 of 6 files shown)


tizianoGuadagnino:
This now potentially invalidates #360, but I feel it is probably what we want.

benemer:
I am a bit confused now, because we now have 3 or even 4 downsampling strategies with corresponding PRs :laughing: 

How about first unifying the Voxel indexing by having one and only one `PointToVoxel`  from the `VoxelHashMap`? We can still discuss together how we want to do the downsampling.

nachovizzo:
For the sake of records, once we also did this. But we decided to split it to avoid inter dependencies between the modules. With this change of I want to compile the preprocessing module I first need the voxel hash map. Additionally, we also wanted to leave the door open for people who might want to borrow the solely one module and thus most of the pieces are isolated (from a build perspective).

That said, we might revise this idea now 💡 as I don't think it was a the most use case of the project.

We could do this but I would avoid this "hacky" thing, as if one reads it is not clear why you are creating a "map"

If we want to merge the implementations I see 2 options

1) extract the voxelization utility into a separate module (hashing  point to voxel, etc) I think we even did this moons ago 🤔
2) create a static method in the voxel hash map that can be used without creating an unnecessary object of the hash map class

Wdyt?

cpp/kiss_icp/core/VoxelUtils.hpp
tizianoGuadagnino, line 1: What about this one?

cpp/kiss_icp/core/VoxelUtils.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace kiss_icp {
+using Voxel = Eigen::Vector3i;
+inline Voxel PointToVoxel(const Eigen::Vector3d &point, const double voxel_size) {
+    return Voxel(static_cast<int>(std::floor(point.x() / voxel_size)),
+                 static_cast<int>(std::floor(point.y() / voxel_size)),
+                 static_cast<int>(std::floor(point.z() / voxel_size)));
+}
+}  // namespace kiss_icp
+
+template <>
```
nachovizzo, line 38: I like the `VoxelHash` operator more as it's slightly more verbose of what's going on under the hood. If you have access to the tsl_robin map, you can see where the hashing function is coming from.  With this partial template specialization, it feels like we are "hiding" a bit the details of our implementation

Remember, we also had it like this in the past and then decided to make it more explicit.

cpp/kiss_icp/core/VoxelUtils.hpp
```diff
@@ -0,0 +1,44 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+#pragma once
+
+#include <Eigen/Core>
+#include <memory>
+
```
nachovizzo, line 28:
```suggestion
#include <cmath>

```

for `std::floor`

cpp/kiss_icp/core/VoxelUtils.hpp
nachovizzo: Then we take option 1.
> extract the voxelization utility into a separate module (hashing point to voxel, etc) I think we even did this moons ago 🤔

Now it's a matter of style-text. I do like this idea; the expense is that now `VoxelHashMap` is not that easy to copy-paste, but I'm happy to let this go.

@benemer what's your view on this?

cpp/kiss_icp/core/VoxelHashMap.hpp
```diff
@@ -29,12 +29,12 @@
 #include <tsl/robin_map.h>
 
 #include <Eigen/Core>
+#include <kiss_icp/core/VoxelUtils.hpp>
```
nachovizzo, line 32:
```suggestion
#include "VoxelUtils.hpp"
```

Shouldn't we follow the local inclusion style like we do with the rest of the code? and also `<>` -> `""`

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -75,7 +75,7 @@ void VoxelHashMap::Update(const std::vector<Eigen::Vector3d> &points, const Soph
 
 void VoxelHashMap::AddPoints(const std::vector<Eigen::Vector3d> &points) {
     std::for_each(points.cbegin(), points.cend(), [&](const auto &point) {
-        auto voxel = PointToVoxel(point);
+        auto voxel = PointToVoxel(point, voxel_size_);
```
nachovizzo, line 78: This is great, as its more explicit. `PointToVoxel(point)` sounds like you magically convert a point to a voxel; without the explicit `voxel_size,`, this doesn't make sense to read. :clap:

cpp/kiss_icp/core/Preprocessing.hpp
```diff
@@ -39,7 +42,4 @@ std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &fram
 /// Originally introduced the calibration factor)
 std::vector<Eigen::Vector3d> CorrectKITTIScan(const std::vector<Eigen::Vector3d> &frame);
 
-/// Voxelize point cloud keeping the original coordinates
```
nachovizzo, line 42: Why did you change the place? I have nothing against it, just makes the diff larger

cpp/kiss_icp/core/Preprocessing.cpp
```diff
@@ -28,42 +28,22 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <cmath>
+#include <kiss_icp/core/VoxelUtils.hpp>
```
nachovizzo, line 31:
```suggestion
#include "VoxelUtils.hpp"
```

Same as before

cpp/kiss_icp/core/Preprocessing.cpp
```diff
@@ -28,42 +28,22 @@
...
-        if (grid.contains(voxel)) continue;
-        grid.insert({voxel, point});
-    }
-    std::vector<Eigen::Vector3d> frame_dowsampled;
-    frame_dowsampled.reserve(grid.size());
-    for (const auto &[voxel, point] : grid) {
-        (void)voxel;
-        frame_dowsampled.emplace_back(point);
-    }
+        if (!grid.contains(voxel)) grid.insert({voxel, point});
```
nachovizzo, line 42: I still prefer the two lines as they are easier to read and do the same. Now, when I read this:

If grid doesn't contain the voxel then insert new `{voxel,  point}` in grid.

Again, it's a language situation, let's @benemer decide which one to vote

cpp/kiss_icp/core/VoxelUtils.hpp
benemer: Yes, I like it. Thanks!

benemer (dismissed):

cpp/kiss_icp/core/Preprocessing.cpp
```diff
@@ -28,42 +28,22 @@
...
+    });
+    std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
```
nachovizzo, line 44: This is not the same as before, and I believe it's slower. This will call the default constructor of `Eigen::Vector3d`, `grid.size()` times, whereas before, we just reserved memory.

Additionally, this is not a transformation operation, and it reads super confusing to me.

That being said if we decide to stick to `std::transform`, then we should use constant iterators `cbegin`, `cend`

EDIT: It's slower, see Quickbench [here](https://quick-bench.com/q/Y7ZmTonpYfD7EpY8iYNYHoqWM4I)
image

nachovizzo (changes requested):

cpp/kiss_icp/core/VoxelUtils.hpp
tizianoGuadagnino: I honestly like this version a bit more, as it uses standard C++ library tools instead of a custom type. Maybe my mind changes over time, also here I will evocate @benemer for the tie.

cpp/kiss_icp/core/Preprocessing.hpp
tizianoGuadagnino: PTSD, the order in the ```.hpp``` and ```.cpp``` was different, it drive me nuts, not it is consistent.

cpp/kiss_icp/core/VoxelUtils.hpp
tizianoGuadagnino: The other advantage is that if somebody just includes this header, ```std::unordered_set```, ```robin_set```, etc., it will work automagically without even specifying the hash function, which is what a user wants in the end.

cpp/kiss_icp/core/VoxelUtils.hpp
tizianoGuadagnino: You right bruda

cpp/kiss_icp/core/Preprocessing.cpp
tizianoGuadagnino: I had a (potentially) better idea using move semantics; I can show you the results here.

image


This is the corresponding [Quickbench](https://quick-bench.com/q/Bx2WNJkfPctQkFykvob0nKRrrAk)

cpp/kiss_icp/core/Preprocessing.hpp
nachovizzo: All great, I'd prefer to leave style changes out of the scope of core changes; when we look back to this PR, it would be better to know the exact real change and not mixin stuff we want to fix.

cpp/kiss_icp/core/Preprocessing.cpp
nachovizzo: And indeed, you are seeing a faster execution because you removed the `static` from the test data generation, if you put it back you will see a slower execution on all this move things

tizianoGuadagnino: Why it was ```static``` in the first place?

tizianoGuadagnino: Mainly because the results change a lot based on that I am not sure which one is more efficient at this point; maybe the ```std::make_move_iterator```is too much for new users, but also the cast to void in the middle of the loop it is not much better, as it is hugely as fuck, if you have a strong opinion for the ```emplace_back``` instead of the new version (I consider the benchmark results inconclusive, as in reality we don't use a ```static``` data structure, I will rather trust my version of the benchmark) I will make it at least a ```for_each```

tizianoGuadagnino: For me, any of the two is fine; the current version is just ugly period

nachovizzo: So, the static was a quick way to share the data among benchmarks that of course didn't work. So forget about quick bench for this situation. I wrote myself a new benchmark and indeed the results points to the direction that the execution is all the same,  so it's a question of readability, not performance.
image

```cpp
#include <benchmark/benchmark.h>

#include <Eigen/Core>
#include <algorithm>
#include <iostream>
#include <memory>
#include <random>
#include <vector>

// Global shared data
std::shared_ptr<const std::vector<Eigen::Vector3d>> shared_data = nullptr;

void InitializeSharedData() {
    if (!shared_data) {
        std::cerr << ">>>>>>>>>>>>>>>>> Creating Shared Data <<<<<<<<<<<<<<<<<" << std::endl;
        std::vector<Eigen::Vector3d> data;

        // Random number generation
        std::random_device rd;
        std::mt19937 gen(rd());
        std::uniform_real_distribution<> dis(-100.0, 100.0);

        for (int i = 0; i < 1 << 20; ++i) {
            double x = dis(gen);
            double y = dis(gen);
            double z = dis(gen);
            data.push_back({x, y, z});
        }

        shared_data = std::make_shared<const std::vector<Eigen::Vector3d>>(std::move(data));
    }
}

static void ReserveEmplaceBack(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> grid = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled;
        frame_dowsampled.reserve(grid.size());
        for (const auto point : grid) {
            frame_dowsampled.emplace_back(point);
        }
        benchmark::DoNotOptimize(frame_dowsampled);
    }
}
static void WithMove(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> grid = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
        std::transform(grid.begin(), grid.end(), frame_dowsampled.begin(),
                       [](auto &point) { return point; });
        benchmark::DoNotOptimize(frame_dowsampled);
    }
}

static void StdCopy(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> grid = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
        std::copy(grid.cbegin(), grid.cend(), frame_dowsampled.begin());
        benchmark::DoNotOptimize(frame_dowsampled);
    }
}

static void StdCopyMove(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> grid = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
        std::copy(std::make_move_iterator(grid.cbegin()), std::make_move_iterator(grid.cend()),
                  frame_dowsampled.begin());
        benchmark::DoNotOptimize(frame_dowsampled);
    }
}

static void PlainMove(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> grid = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
        std::move(grid.cbegin(), grid.cend(), frame_dowsampled.begin());
        benchmark::DoNotOptimize(frame_dowsampled);
    }
}

BENCHMARK(ReserveEmplaceBack)->Unit(benchmark::kMillisecond);
BENCHMARK(WithMove)->Unit(benchmark::kMillisecond);
BENCHMARK(StdCopy)->Unit(benchmark::kMillisecond);
BENCHMARK(StdCopyMove)->Unit(benchmark::kMillisecond);
BENCHMARK(PlainMove)->Unit(benchmark::kMillisecond);

int main(int argc, char **argv) {
    InitializeSharedData();
    ::benchmark::Initialize(&argc, argv);
    ::benchmark::RunSpecifiedBenchmarks();
}
```

tizianoGuadagnino: Actually, as always, STL for the win on everything 

image

The improvement I saw is due to STL algorithm optimization rather than the move logic itself. 

As always thanks to all the gods for STL.

tizianoGuadagnino: For reference [this](https://quick-bench.com/q/PKnGsB9xDjYHn8L95_O_2anDI_o) is the benchmark

tizianoGuadagnino: Adding ```static``` again...

image

nachovizzo: Actually not, I suggest we stop using this quick bench garbage as it gives contracictiorady results. I tested this locally, and the runtime is literally the same.

image

```cpp
#include <benchmark/benchmark.h>
#include <tsl/robin_map.h>

#include <Eigen/Core>
#include <algorithm>
#include <iostream>
#include <memory>
#include <random>
#include <vector>

#include "VoxelUtils.hpp"
using namespace kiss_icp;

// Global shared data
std::shared_ptr<const std::vector<Eigen::Vector3d>> shared_data = nullptr;

void InitializeSharedData() {
    if (!shared_data) {
        std::cerr << ">>>>>>>>>>>>>>>>> Creating Shared Data <<<<<<<<<<<<<<<<<" << std::endl;
        std::vector<Eigen::Vector3d> data;

        // Random number generation
        std::random_device rd;
        std::mt19937 gen(rd());
        std::uniform_real_distribution<> dis(-100.0, 100.0);

        for (int i = 0; i < 1 << 10; ++i) {
            double x = dis(gen);
            double y = dis(gen);
            double z = dis(gen);
            data.push_back({x, y, z});
        }

        shared_data = std::make_shared<const std::vector<Eigen::Vector3d>>(std::move(data));
    }
}

std::vector<Eigen::Vector3d> VoxelDownsampleNew(const std::vector<Eigen::Vector3d> &frame,
                                                double voxel_size) {
    tsl::robin_map<Voxel, Eigen::Vector3d> grid;
    grid.reserve(frame.size());
    std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
        const auto voxel = PointToVoxel(point, voxel_size);
        if (!grid.contains(voxel)) grid.insert({voxel, point});
    });

    std::vector<Eigen::Vector3d> frame_dowsampled;
    frame_dowsampled.reserve(grid.size());
    for (const auto &voxel_and_point : grid) frame_dowsampled.emplace_back(voxel_and_point.second);

    return frame_dowsampled;
}

std::vector<Eigen::Vector3d> VoxelDownsampleOld(const std::vector<Eigen::Vector3d> &frame,
                                                double voxel_size) {
    tsl::robin_map<Voxel, Eigen::Vector3d> grid;
    grid.reserve(frame.size());
    std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
        const auto voxel = PointToVoxel(point, voxel_size);
        if (!grid.contains(voxel)) grid.insert({voxel, point});
    });

    std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
    std::transform(grid.cbegin(), grid.cend(), frame_dowsampled.begin(),
                   [](const auto &voxel_and_point) { return voxel_and_point.second; });

    return frame_dowsampled;
}

std::vector<Eigen::Vector3d> VoxelDownsampleForEach(const std::vector<Eigen::Vector3d> &frame,
                                                    double voxel_size) {
    tsl::robin_map<Voxel, Eigen::Vector3d> grid;
    grid.reserve(frame.size());
    std::for_each(frame.cbegin(), frame.cend(), [&](const auto &point) {
        const auto voxel = PointToVoxel(point, voxel_size);
        if (!grid.contains(voxel)) grid.insert({voxel, point});
    });

    std::vector<Eigen::Vector3d> frame_dowsampled;
    frame_dowsampled.reserve(grid.size());
    std::for_each(grid.cbegin(), grid.cend(), [&](const auto &voxel_and_point) {
        frame_dowsampled.emplace_back(voxel_and_point.second);
    });

    return frame_dowsampled;
}

static void NewVoxelDownsample(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> frame = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled = VoxelDownsampleNew(frame, 1.0);
    }
}

static void OldVoxelDownsample(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> frame = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled = VoxelDownsampleOld(frame, 1.0);
    }
}
static void ForEach(benchmark::State &state) {
    for (auto _ : state) {
        std::vector<Eigen::Vector3d> frame = *shared_data;
        std::vector<Eigen::Vector3d> frame_dowsampled = VoxelDownsampleForEach(frame, 1.0);
    }
}

BENCHMARK(NewVoxelDownsample);
BENCHMARK(OldVoxelDownsample);
BENCHMARK(ForEach);

int main(int argc, char **argv) {
    InitializeSharedData();
    ::benchmark::Initialize(&argc, argv);
    ::benchmark::RunSpecifiedBenchmarks();
}
```

tizianoGuadagnino: Make sense now, to confirm, I was testing on real data:

### **main**

image

### **transform**

image

### **for_each**

image

tizianoGuadagnino: Literally same, shall we go for the ```for_each```?

cpp/kiss_icp/core/Preprocessing.cpp
```diff
@@ -43,9 +43,11 @@ std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d>
         const auto voxel = PointToVoxel(point, voxel_size);
         if (!grid.contains(voxel)) grid.insert({voxel, point});
     });
-    std::vector<Eigen::Vector3d> frame_dowsampled(grid.size());
-    std::transform(grid.cbegin(), grid.cend(), frame_dowsampled.begin(),
-                   [](const auto &voxel_and_point) { return voxel_and_point.second; });
+    std::vector<Eigen::Vector3d> frame_dowsampled;
+    frame_dowsampled.reserve(grid.size());
+    std::for_each(grid.cbegin(), grid.cend(), [&](const auto &voxel_and_point) {
```
nachovizzo, line 47: @tizianoGuadagnino commenting here as the other thread is too large now

I think, overall, this is the best approach, as we mainly use for_each everywhere in the code. It might be too much for newbies, but this also forces them to read what is the algo doing, as it's now everywhere in the code

Let's keep this double for each implementation, is short and easier to read than `transform, make, my mandess`

cpp/kiss_icp/core/Preprocessing.cpp
nachovizzo: Yes!

cpp/kiss_icp/core/Preprocessing.cpp
tizianoGuadagnino: Awesome, I really like this new version.

nachovizzo:
@tizianoGuadagnino is it too much to ask now for splitting this into 2 PR's? 90% of the changes are style changes and not the original idea we had for this PR?

I will like to have 2 commits in the history, 1 for the moving the existing utilities to a separate header, and a follow up changing style, raw for loops for for each, etc

tizianoGuadagnino:
> @tizianoGuadagnino is it too much to ask now for splitting this into 2 PR's? 90% of the changes are style changes and not the original idea we had for this PR?
> 
> I will like to have 2 commits in the history, 1 for the moving the existing utilities to a separate header, and a follow up changing style, raw for loops for for each, etc

I can, but spiritually, these two changes are actually one, which is **_we revise everything that regards Voxels_**. This includes ```VoxelUtils```, ```VoxelHash```, ```VoxelHashMap``` and ```Voxeldownsample```. Or maybe you mean we do a separate PR for ```for``` to ```std::for_each```?. If you have a strong opinion of this, I can do it, but can I consider all the changes accepted? 

In the meantime, I will change the name of the PR to be consistent.

cpp/kiss_icp/core/Preprocessing.hpp
tizianoGuadagnino, line 1: Here I put back ```VoxelDownsample``` to the original location to don't pollute the diff here.

nachovizzo:
@tizianoGuadagnino I updated this PR with the only changes that should be, without any style change

Let's follow up the style changes in https://github.com/PRBonn/kiss-icp/pull/363/files

nachovizzo (approved):

cpp/kiss_icp/core/VoxelUtils.hpp
nachovizzo: Moving this discussion to #364

benemer (approved):
Awesome, thank you guys :)

PR #364, opened 2024-07-11
Style changes in voxelization and preprocessing utilities

nachovizzo:
Continuation of #362 

Style changes on voxelization utilities:

- Move the voxel downsample function so it belongs to its family (VoxelUtils)
- minor code style changes in voxelization
- Remove `VoxelHash` struct favoring partial template specialization of `std::hash` so the end user doesn't care what's going on behind the scenes

### DIFF
--- a/cpp/kiss_icp/core/Preprocessing.hpp
+++ b/cpp/kiss_icp/core/Preprocessing.hpp
@@ -30,7 +30,7 @@
 namespace kiss_icp {
 /// Voxelize point cloud keeping the original coordinates
 std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
-                                             double voxel_size);
+                                             const double voxel_size);
 
 /// Crop the frame with max/min ranges
 std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &frame,
--- a/cpp/kiss_icp/core/Preprocessing.hpp
+++ b/cpp/kiss_icp/core/Preprocessing.hpp
@@ -28,9 +28,6 @@
 #include <vector>
 
 namespace kiss_icp {
-/// Voxelize point cloud keeping the original coordinates
-std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
-                                             const double voxel_size);
 
 /// Crop the frame with max/min ranges
 std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &frame,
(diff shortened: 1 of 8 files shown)


cpp/kiss_icp/core/Preprocessing.hpp
```diff
@@ -28,6 +28,9 @@
 #include <vector>
 
 namespace kiss_icp {
+/// Voxelize point cloud keeping the original coordinates
```
nachovizzo, line 31: One question to discuss now in this particular PR:

Assuming we get rid of `CorrectKITTIScan`, then this "Preprocessing" module will only have two functions:

```cpp
std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &frame,
                                             const double voxel_size);

std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &frame,
                                        double max_range,
                                        double min_range);
```

The `Preprocess` function, which is only cropping (so we could change the name of it), is only consumed in the kiss-icp pipeline, so potentially, we could leave it there.

This makes me think `VoxelDownsample` is the only missing function here, and we can put it in our new and shiny `VoxelUtiils.hpp` module.

cpp/kiss_icp/core/VoxelUtils.hpp
```diff
@@ -25,21 +25,25 @@
...
 }
+/// Voxelize a point cloud keeping the original coordinates
+std::vector<Eigen::Vector3d> VoxelDownsample(const std::vector<Eigen::Vector3d> &point_cloud,
+                                             const double voxel_size);
 
-struct VoxelHash {
-    size_t operator()(const Voxel &voxel) const {
+}  // namespace kiss_icp
+
+template <>
```
nachovizzo, line 43: so we need to vote for partial template specialisation or custom type, it's 1-1 now, so it's @benemer choice now :)

Follow up of https://github.com/PRBonn/kiss-icp/pull/362#discussion_r1673016356

cpp/kiss_icp/core/Preprocessing.hpp
nachovizzo: Moving this to a separate PR now https://github.com/PRBonn/kiss-icp/pull/369

cpp/kiss_icp/core/Preprocessing.hpp
```diff
@@ -39,7 +38,4 @@ std::vector<Eigen::Vector3d> Preprocess(const std::vector<Eigen::Vector3d> &fram
 /// Originally introduced the calibration factor)
 std::vector<Eigen::Vector3d> CorrectKITTIScan(const std::vector<Eigen::Vector3d> &frame);
 
-/// Voxelize point cloud keeping the original coordinates
```
nachovizzo, line 42: we should also remove it then from the .cpp

cpp/kiss_icp/core/VoxelUtils.hpp
benemer: I agree with @nachovizzo that using the `VoxelHash` struct makes reading a bit easier. However, I am not sure how "accessible" this hash function is in the first place. For example, it is not clear where it is coming from and what those numbers are. Given this, I think I am okay with "hiding" it and using the partial specialization.

nachovizzo: oki doki!

nachovizzo:
Now ready to review and :ship:

benemer (approved):
Can be merged from my side :+1:

tizianoGuadagnino (approved):

PR #365, opened 2024-07-11
Fix include headers

saurabh1002:
As a good student of @nachovizzo , I propose fixes to be consistent with the [IWYU](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md) principle.

### DIFF
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -26,9 +26,9 @@
 #include <tsl/robin_map.h>
 
 #include <Eigen/Core>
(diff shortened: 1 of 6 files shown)


saurabh1002:
image

```Eigen::AxisAngled``` is defined in `Eigen/Geometry`, which was getting included through `Sophus/se3.hpp` even when **Preprocessing.cpp** and **Preprocessing.hpp** made no use of the Sophus library.

tizianoGuadagnino (approved):

nachovizzo (approved):
Thanks guptu

PR #373, opened 2024-07-24
Remove unncessary python list appends

saurabh1002:
1. Python lists and list appends were used to store odomtery poses and execution times
2. But we already know the expected length of these variables
3. So, we can use numpy arrays directly

### DIFF
--- a/python/kiss_icp/pipeline.py
+++ b/python/kiss_icp/pipeline.py
@@ -99,7 +99,9 @@ class OdometryPipeline:
             source, keypoints = self.odometry.register_frame(raw_frame, timestamps)
             self.poses[idx - self._first] = self.odometry.last_pose
             self.times[idx - self._first] = time.perf_counter_ns() - start_time
-            self.visualizer.update(source, keypoints, self.odometry.local_map, self.odometry.last_pose)
+            self.visualizer.update(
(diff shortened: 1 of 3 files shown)


python/kiss_icp/pipeline.py
```diff
@@ -97,8 +97,8 @@ def _run_pipeline(self):
             raw_frame, timestamps = self._next(idx)
             start_time = time.perf_counter_ns()
             source, keypoints = self.odometry.register_frame(raw_frame, timestamps)
-            self.poses.append(self.odometry.last_pose)
-            self.times.append(time.perf_counter_ns() - start_time)
+            self.poses[idx - self._first] = self.odometry.last_pose
+            self.times[idx - self._first] = time.perf_counter_ns() - start_time
             self.visualizer.update(source, keypoints, self.odometry.local_map, self.poses[-1])
```
benemer, line 102: This currently breaks the visualizer, because the last pose is initialized with zeros, we can use `self.odometry.last_pose` here

benemer (changes requested):
Can you directly open a branch on this repo next time? In that case we can change minor things ourselves like this visualizer update :pray:

benemer (approved):

tizianoGuadagnino (approved):

PR #377, opened 2024-07-26
Fix EXISTS resolves symlinks but add_subdirectory does not

MCFurry:
Our tooling typically clones projects in symlinked directories where the `EXISTS` and `add_subdirectory` logic doesn't work out of the box.
This way it does!

### DIFF
--- a/ros/CMakeLists.txt
+++ b/ros/CMakeLists.txt
@@ -25,7 +25,7 @@ project(kiss_icp VERSION 1.0.0 LANGUAGES CXX)
 
 set(CMAKE_BUILD_TYPE Release)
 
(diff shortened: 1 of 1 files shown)


nachovizzo (approved):
Looks good to me! Any suggestion on this build logic is highly appreciated. We mainly wanted to leave the C++ ROS-clean, and this is the way we find to store two monsters in one repository. But we were never SUPER happy about this hacky subdirectory injection

MCFurry:
I see, I guess this is a way to do it.
Here is another example: https://github.com/nobleo/xsens_mti_driver where a clean lib is build together with a ROS wrapper.

Another option would be to indeed create two repositories where one is solely the library and another the ROS wrapper/node.

If you appreciate it, we can sketch up a PR with one of these options?

benemer (approved):

PR #379, opened 2024-07-30
Kissualizer: a new KISS-ICP visualizer!

l00p3:
I want to propose a new visualizer for KISS-ICP.  The current visualizer presents some issues in my opinion:
- No GUI: this reduces a lot the usability, one have to read the keyboard key to press on the terminal in order to move around
- The previous point become even more problematic because of Open3D warnings that appear sometime
- Passing from global view to local view (and the other way around) requires to trigger the next frame in order to visualize the changes
- No possibility to modify point size individually for each cloud visualized
- No possibility to visualize any other information, because, again, no GUI
- No flexibility at all in case ones want to introduce new features
- Copies happening in the code, due to the incompatibility between data structures used

All these cons are there because Open3D is not meant to be used as a general visualizer.

I'll take over here: Kissualizer, a new visualizer using [polyscope](https://polyscope.run/py/), a library which purpose is exactly visualize 3D data. Here some pros presented with my visualizer:
- GUI: we can use graphical buttons, slides and whatever to control our visualization, still maintaining the possibility to use the keyboard (also the possibility to add a visual legend on the screen opens here)
- No weird warning appearing in the terminal
- Global/Local views switch is not anymore a problem
- Point sizes can be modified individually for each cloud
- Possibility to visualize other information, like FPS, current parameters used and so on
- Background color change with palette
- Huge flexibility for the introduction of new features
- No copy happening in the background, polyscope works with the data structures that you already use

This is just to scratch the surface, with polyscope you can give vent to all your dirtiest graphic fantasies!

Here a demo to show the current implementation that I propose (the frame rate seems pretty low because of the GIF format, try it on your machine!):

KissualizerDemo

Of course I just added some basic features, but this can be really expanded a lot. For now I just want to show you the potential of using a real visualizer. In case you like it I can work on new features, better implementation code side and whatever you want. Also, I used a slighlty different color palette, but this, of course, can be adjusted.
It is time for KISS-ICP to embrace the future!

[Here](https://github.com/nmwsharp/polyscope) the github page of polyscope in case you need more information. [Here](https://github.com/nmwsharp/polyscope-py) the 'pybinded' version.

Oh, this would add only a simple optional dependency python side, a simple ```pip install polyscope```.


nachovizzo:
Waaaaaaaaaat 🤤 I dream finally come true 🥹 I always wanted to use the new visualizer and leave the legacy open3d behind

Thanks!!!!!!! I'll check this out as soon as possible

nachovizzo:
@benemer , @tizianoGuadagnino 

Do we want to keep a legacy visualizer for those who "don't want / can't" install polyscope? I checked the compatibility of `polyscope` with different platforms/os, and it's on par with our compatibility check, so it shouldn't be a problem. Also, its only binary dependency is numpy.

I vote to kill the old visualizer but just raising this concern here in case someone else have a different opinion

tizianoGuadagnino:
@nachovizzo for me the old visualizer should die, in principle it should never being born. It is fair to kill it now

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
 
-CYAN = np.array([0.24, 0.898, 1])
-RED = np.array([128, 0, 0]) / 255.0
-BLACK = np.array([0, 0, 0]) / 255.0
-BLUE = np.array([0.4, 0.5, 0.9])
-GRAY = np.array([0.4, 0.4, 0.4])
-SPHERE_SIZE = 0.20
+from kiss_icp.config.parser import KISSConfig
+
+BACKGROUND_COLOR = [0.8470, 0.8588, 0.8863]
```
nachovizzo, line 31: I'm not a big fan of discussing colors, mainly due to my condition. But I would probably pick the color palette we had before, just to make it look consistent with the ROS visualizations, that we make them look the same in https://github.com/PRBonn/kiss-icp/pull/285

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
-class RegistrationVisualizer(StubVisualizer):
-    # Public Interaface ----------------------------------------------------------------------------
-    def __init__(self):
-        try:
-            self.o3d = importlib.import_module("open3d")
-        except ModuleNotFoundError as err:
-            print(f'open3d is not installed on your system, run "pip install open3d"')
-            exit(1)
+def start_pause_callback():
+    button_name = "PAUSE" if Kissualizer.play_mode else "START"
```
nachovizzo, line 50: Not sure if this is the render, but should we add
`START [Space]` and things like that, now the keyboard still works, which I love, as I'm not a mouse guy, but the letters for command are hidden.

The other option would be to spit in the user face the commands, like before

l00p3: I was thinking the same. Probably I can add a section in the GUI with the list of commands, or just add the key "on the buttons". I'll test both and post a screenshot.

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
+from kiss_icp.config.parser import KISSConfig
+
+BACKGROUND_COLOR = [0.8470, 0.8588, 0.8863]
+FRAME_COLOR = [0.2156, 0.2471, 0.3176]
+KEYPOINTS_COLOR = [0.8588, 0.3294, 0.3803]
+LOCAL_MAP_COLOR = [0.3451, 0.6431, 0.6902]
+TRAJECTORY_COLOR = [0.1058, 0.1059, 0.1176]
+FRAME_PTS_SIZE = 0.2
+KEYPOINTS_PTS_SIZE = 0.3
+MAP_PTS_SIZE = 0.1
```
benemer, line 38:
```suggestion
BACKGROUND_COLOR = [0.0, 0.0, 0.0]
FRAME_COLOR = [0.4, 0.5, 0.9]
KEYPOINTS_COLOR = [0.24, 0.898, 1]
LOCAL_MAP_COLOR = [0.4, 0.4, 0.4]
TRAJECTORY_COLOR = [0.1, 0.1, 0.1]
FRAME_PTS_SIZE = 0.1
KEYPOINTS_PTS_SIZE = 0.1
MAP_PTS_SIZE = 0.1
```

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
 
 
 class StubVisualizer(ABC):
-    def __init__(self):
+    def __init__(self, config: KISSConfig):
         pass
 
-    def update(self, source, keypoints, target_map, pose):
+    def update(self, source, keypoints, target_map, pose, lest_time):
```
benemer, line 45:
```suggestion
    def update(self, source, keypoints, target_map, pose, last_time):
```

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
 
-class RegistrationVisualizer(StubVisualizer):
-    # Public Interaface ----------------------------------------------------------------------------
-    def __init__(self):
-        try:
-            self.o3d = importlib.import_module("open3d")
-        except ModuleNotFoundError as err:
-            print(f'open3d is not installed on your system, run "pip install open3d"')
-            exit(1)
+def start_pause_callback():
```
benemer, line 49: Why aren't these methods part of the Kissualizer class?

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
-        # Cache the state of the visualizer
-        self.state = (
-            self.render_map,
-            self.render_keypoints,
-            self.render_source,
-        )
+def screenshot_callback():
+    # TODO: this is just for demo, set a more valid path
+    if Kissualizer.polyscope.imgui.Button("SCREENSHOT"):
+        image_filename = "screenshot.jpg"
```
benemer, line 67: How about using the current date and time in the filename? The path is fine for me. Otherwise we can also save it to the results folder.

python/kiss_icp/tools/visualizer.py
```diff
@@ -20,223 +20,301 @@
...
-            "\t    [M] to toggle on/off the local map\n"
-            "\t    [V] to toggle ego/global viewpoint\n"
-            "\t    [T] to toggle the trajectory view(only available in global view)\n"
-            "\t    [C] to center the viewpoint\n"
-            "\t    [W] to toggle a white background\n"
-            "\t    [B] to toggle a black background\n"
-        )
+def fps_callback():
+    if Kissualizer.play_mode:
+        total_time_s = np.sum(Kissualizer.times) * 1e-9
```
benemer, line 74: This duplicates the fps computation from the pipeline, and the timestamps are not used anywhere else in the visualizer. How about making the `get_fps` method from the pipeline a member and only passing the fps to the visualizer?

benemer: Or we generalize this and make it possible to pass a dict of additional data to the visualizer. This callback will then print sequentially all keys and values.

benemer (changes requested):
Awesome, I love it. Thank you for this great work! :partying_face: 

I found a few minor points. For some reason, the frame color is still different from the one we have in the old visualizer.

Also, I am not sure if I like the old trajectory more or this one. Because with the old one, you could see how many frames you estimated in a local area and you could get an idea of the acceleration of the platform.

What do you think?

benemer:
> @benemer , @tizianoGuadagnino
> 
> Do we want to keep a legacy visualizer for those who "don't want / can't" install polyscope? I checked the compatibility of `polyscope` with different platforms/os, and it's on par with our compatibility check, so it shouldn't be a problem. Also, its only binary dependency is numpy.
> 
> I vote to kill the old visualizer but just raising this concern here in case someone else have a different opinion

Yes, let's kill it. If you can afford to install open3d, you can also install polyscope instead, right?

python/kiss_icp/tools/visualizer.py
l00p3: Screenshot from 2024-07-31 16-59-20
This is how it appears now, I kinda like it. Let me know what do you think

python/kiss_icp/tools/visualizer.py
l00p3: I like the "dict" idea, I will do it!

python/kiss_icp/tools/visualizer.py
l00p3: I was thinking about the results folder, probably better, just the path need to be computed in the pipeline also. I'll try to do it

python/kiss_icp/tools/visualizer.py
l00p3: To avoid to do them all static, because the gui is a "global" entity and takes only callbacks without any argument. If they become Kissualizer's methods then we have to pass "self" as parameter.

python/kiss_icp/tools/visualizer.py
nachovizzo: fantastic

nachovizzo (approved):
Fantastic, a follow up in `lidar-visualizer` would also be nice right?

tizianoGuadagnino (dismissed):
I fix the color palette and point size in such a way that everything is properly visible, if anyone have anything to say, talk now or die forever.

tizianoGuadagnino (dismissed):

l00p3:
Ok, I think now the shape is way better. if you have other requests I am here to satisfy them.

l00p3:
Just another feature added: if you are in global view and in PAUSE mode, you can double click on a pose on the trajectory to visualize its position in the global frame. Let me know if you want to keep this feature. 
I want to point out the fact that this "click" on object to visualize some related information opens a lot of possibilities if you want. Good to keep in mind.

benemer (approved):
Awesome, I love it! Thanks for all the effort :heart:

nachovizzo (approved):

PR #383, opened 2024-08-05
Add draft visualization for the voxels of the VoxelHashMap

nachovizzo:
WIP

This is a new idea to visualize the internal map representation of kiss icp besides the pointcloud stored in the hash table.
@l00p3 do you think something like this would be possible with the new visualizer?

The current Open3D API for `VoxelGrid` is not that rich, and it makes rendering a bit slow when the grid is requested, so I'd be more than happy to migrate so something better.

image

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -58,17 +58,9 @@ std::vector<Eigen::Vector3d> VoxelHashMap::Pointcloud() const {
 
 // Get the indices of the occupided voxels as points, mainly used for visualization
 std::vector<Voxel> VoxelHashMap::GetVoxels() const {
-std::vector<Voxel> voxels(map_.size());
+    std::vector<Voxel> voxels(map_.size());
     std::transform(map_.cbegin(), map_.cend(), voxels.begin(),
                    [](const auto &map_element) { return map_element.first; });
-    return voxels;
-}
-    std::vector<Voxel> voxels;
-    voxels.reserve(map_.size());
-    for (const auto &[voxel, voxel_block] : map_) {
-        (void)voxel_block;
-        voxels.emplace_back(voxel);
-    }
     return voxels;
 }
 
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -56,6 +56,17 @@ std::vector<Eigen::Vector3d> VoxelHashMap::Pointcloud() const {
     return points;
 }
 
+// Get the indices of the occupided voxels as points, mainly used for visualization
+std::vector<Voxel> VoxelHashMap::GetVoxels() const {
+    std::vector<Voxel> voxels;
+    voxels.reserve(map_.size());
+    for (const auto &[voxel, voxel_block] : map_) {
+        (void)voxel_block;
+        voxels.emplace_back(voxel);
+    }
+    return voxels;
+}
+
 void VoxelHashMap::Update(const std::vector<Eigen::Vector3d> &points,
                           const Eigen::Vector3d &origin) {
     AddPoints(points);
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -58,6 +58,11 @@ std::vector<Eigen::Vector3d> VoxelHashMap::Pointcloud() const {
 
 // Get the indices of the occupided voxels as points, mainly used for visualization
 std::vector<Voxel> VoxelHashMap::GetVoxels() const {
+std::vector<Voxel> voxels(map_.size());
+    std::transform(map_.cbegin(), map_.cend(), voxels.begin(),
+                   [](const auto &map_element) { return map_element.first; });
+    return voxels;
+}
     std::vector<Voxel> voxels;
     voxels.reserve(map_.size());
     for (const auto &[voxel, voxel_block] : map_) {
--- a/python/kiss_icp/tools/visualizer.py
+++ b/python/kiss_icp/tools/visualizer.py
@@ -34,6 +34,7 @@ RED = np.array([128, 0, 0]) / 255.0
 BLACK = np.array([0, 0, 0]) / 255.0
 BLUE = np.array([0.4, 0.5, 0.9])
 GRAY = np.array([0.4, 0.4, 0.4])
+LIGHT_GRAY = np.array([0.9, 0.9, 0.9])
 SPHERE_SIZE = 0.20
 
 
@@ -63,6 +64,8 @@ class RegistrationVisualizer(StubVisualizer):
         self.source = self.o3d.geometry.PointCloud()
         self.keypoints = self.o3d.geometry.PointCloud()
         self.target = self.o3d.geometry.PointCloud()
+        self.voxel_map = self.o3d.geometry.VoxelGrid()
+        self.voxel_map.voxel_size = 1.0
         self.frames = []
 
         # Initialize visualizer
@@ -74,6 +77,7 @@ class RegistrationVisualizer(StubVisualizer):
         self.render_map = True
         self.render_source = True
         self.render_keypoints = False
+        self.render_voxel_grid = True
         self.global_view = False
         self.render_trajectory = True
         # Cache the state of the visualizer
@@ -81,11 +85,11 @@ class RegistrationVisualizer(StubVisualizer):
             self.render_map,
             self.render_keypoints,
             self.render_source,
+            self.render_voxel_grid,
         )
 
     def update(self, source, keypoints, target_map, pose):
-        target = target_map.point_cloud()
-        self._update_geometries(source, keypoints, target, pose)
+        self._update_geometries(source, keypoints, target_map, pose)
         while self.block_vis:
             self.vis.poll_events()
             self.vis.update_renderer()
@@ -100,8 +104,10 @@ class RegistrationVisualizer(StubVisualizer):
         self.vis.add_geometry(self.source)
         self.vis.add_geometry(self.keypoints)
         self.vis.add_geometry(self.target)
+        self.vis.add_geometry(self.voxel_map)
         self._set_black_background(self.vis)
         self.vis.get_render_option().point_size = 1
+        self.vis.get_render_option().mesh_show_wireframe = True
         print(
             f"{w_name} initialized. Press:\n"
             "\t[SPACE] to pause/start\n"
@@ -130,6 +136,7 @@ class RegistrationVisualizer(StubVisualizer):
         self._register_key_callback(["F"], self._toggle_source)
         self._register_key_callback(["K"], self._toggle_keypoints)
         self._register_key_callback(["M"], self._toggle_map)
+        self._register_key_callback(["G"], self._toggle_voxel_grid)
         self._register_key_callback(["T"], self._toggle_trajectory)
         self._register_key_callback(["B"], self._set_black_background)
         self._register_key_callback(["W"], self._set_white_background)
@@ -172,6 +179,10 @@ class RegistrationVisualizer(StubVisualizer):
         self.render_map = not self.render_map
         return False
 
+    def _toggle_voxel_grid(self, vis):
+        self.render_voxel_grid = not self.render_voxel_grid
+        return False
+
     def _toggle_view(self, vis):
         self.global_view = not self.global_view
         self._trajectory_handle()
@@ -194,7 +205,7 @@ class RegistrationVisualizer(StubVisualizer):
             for frame in self.frames:
                 self.vis.remove_geometry(frame, reset_bounding_box=False)
 
-    def _update_geometries(self, source, keypoints, target, pose):
+    def _update_geometries(self, source, keypoints, target_map, pose):
         # Source hot frame
         if self.render_source:
             self.source.points = self.o3d.utility.Vector3dVector(source)
@@ -215,7 +226,7 @@ class RegistrationVisualizer(StubVisualizer):
 
         # Target Map
         if self.render_map:
-            target = copy.deepcopy(target)
+            target = target_map.point_cloud()
             self.target.points = self.o3d.utility.Vector3dVector(target)
             if self.global_view:
                 self.target.paint_uniform_color(GRAY)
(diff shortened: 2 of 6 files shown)


nachovizzo:
Well, it turns out that #354 was much more than a simple bugfix. ...

 By taking this branch and applying a "revert patch," we can see that a nice bug has been in the pipeline since the beginning.  The following map renders are the same dataset at the same timestamp. Do you see any difference? I used **voxel size of 10 meters** to make the error 100% visible.

## main branch today
Screenshot from 2024-08-05 18-29-57

## Before #354

Please pay attention to all the points that are part of the map but are not associated with the proper voxel, as it's not being rendered. Also, please pay attention to this only happens 100% of the cases. We have a rounding error on the negative quadrants (the center of the screen is the origin, X positive looking right)
image

## What happened?

It turns out that we were computing the **wrong** voxel for some points, associating voxel blocks with points outside the voxel's range.

## How to test
```diff
diff --git a/cpp/kiss_icp/core/VoxelHashMap.cpp b/cpp/kiss_icp/core/VoxelHashMap.cpp
index a5cb6ef..14f0eb9 100644
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -84,7 +84,7 @@ void VoxelHashMap::Update(const std::vector<Eigen::Vector3d> &points, const Soph
 void VoxelHashMap::AddPoints(const std::vector<Eigen::Vector3d> &points) {
     const double map_resolution = std::sqrt(voxel_size_ * voxel_size_ / max_points_per_voxel_);
     std::for_each(points.cbegin(), points.cend(), [&](const auto &point) {
-        auto voxel = PointToVoxel(point, voxel_size_);
+        auto voxel = Voxel((point / voxel_size_).template cast<int>());
         auto search = map_.find(voxel);
         if (search != map_.end()) {
             auto &voxel_points = search.value();
diff --git a/cpp/kiss_icp/core/VoxelUtils.hpp b/cpp/kiss_icp/core/VoxelUtils.hpp
index 580a252..cd38077 100644
--- a/cpp/kiss_icp/core/VoxelUtils.hpp
+++ b/cpp/kiss_icp/core/VoxelUtils.hpp
@@ -31,9 +31,9 @@ namespace kiss_icp {
 
 using Voxel = Eigen::Vector3i;
 inline Voxel PointToVoxel(const Eigen::Vector3d &point, const double voxel_size) {
-    return Voxel(static_cast<int>(std::floor(point.x() / voxel_size)),
-                 static_cast<int>(std::floor(point.y() / voxel_size)),
-                 static_cast<int>(std::floor(point.z() / voxel_size)));
+    return Voxel(static_cast<int>(point.x() / voxel_size),
+                 static_cast<int>(point.y() / voxel_size),
+                 static_cast<int>(point.z() / voxel_size));
 }
 
 /// Voxelize a point cloud keeping the original coordinates
diff --git a/python/kiss_icp/config/config.py b/python/kiss_icp/config/config.py
index cbb9753..ded83e3 100644
--- a/python/kiss_icp/config/config.py
+++ b/python/kiss_icp/config/config.py
@@ -32,7 +32,7 @@ class DataConfig(BaseModel):
 
 
 class MappingConfig(BaseModel):
-    voxel_size: Optional[float] = None  # default: take it from data
+    voxel_size: Optional[float] = 10.0  # default: take it from data
     max_points_per_voxel: int = 20
```
 
## Further sanity checks

I also added this silly code to the `VoxelHashMap::AddPoints` method:

```cpp
      // Sanity check that the point correspond to the voxel
        const double min_x = (voxel_size_ * voxel.x());
        const double max_x = (voxel_size_ * (voxel.x() + 1));

        const double min_y = (voxel_size_ * voxel.y());
        const double max_y = (voxel_size_ * (voxel.y() + 1));

        const double min_z = (voxel_size_ * voxel.z());
        const double max_z = (voxel_size_ * (voxel.z() + 1));

        if (point.x() > max_x || point.x() < min_x || point.y() > max_y || point.y() < min_y ||
            point.z() > max_z || point.z() < min_z

        ) {
            std::cerr << "[ERROR] \n"
                      << "point = " << point.transpose() << std::endl
                      << "voxel = " << voxel.transpose() << std::endl
                      << "min_x: " << min_x << ", max_x: " << max_x << std::endl
                      << "min_y: " << min_y << ", max_y: " << max_y << std::endl
                      << "min_z: " << min_z << ", max_z: " << max_z << std::endl;
        }
```

Mainly to double-check this situation, and indeed, if you try this with the code from before #354, you get lots of errors of points that fell under the wrong voxel and thus, got inserted in the VoxelHashMap but being accessed wrongly (actually not at all)

Example:
```

[ERROR] point= -59.7119 -30.1762  2.86053, voxel = -5 -3  0
point = -49.4642 -57.2834  3.04068, voxel = -1 -1  0
-5488, point =  -78.4972 -0.613767  -1.41267
voxel = -2  0  0
min_x: -60, max_x: -30
min_y: 0, max_y: 30
min_z: 0, max_z: 30
```

## What was the impact of this bug we had:

I don't think it is that severe, but keep in mind that when doing `std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels)`, the voxel we are querying, it contains points that are far away from the bounds of the voxel, and thus (if the adaptive the admits for the case) will introduce wrong data associations in the system

I think it is not to freak out, as we fixed, but it is interesting learning to share with the KISS team

CC @benemer , @tizianoGuadagnino , @saurabh1002 . Please do share your thoughts just to confirm I'm not 100% crazy

PS: This shows the **need** of having the ability to plot the internal voxel hash map structure. We would have discovered this issue much before if we had a way to visualize how those points from the map were stored in our data structure. Cheers

l00p3:
Regarding the visualization, I think the same can be achieved in some way with Polyscope. Probably it will require some massage, but should be faster than the Open3D implementation (I hope). I'll update you as soon as I can.

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -56,6 +56,17 @@ std::vector<Eigen::Vector3d> VoxelHashMap::Pointcloud() const {
     return points;
 }
 
+// Get the indices of the occupided voxels as points, mainly used for visualization
+std::vector<Voxel> VoxelHashMap::GetVoxels() const {
```
l00p3, line 60: I would write this function with a transform, just to be pedantic :) 
```suggestion
std::vector<Voxel> VoxelHashMap::GetVoxels() const {
std::vector<Voxel> voxels(map_.size());
    std::transform(map_.cbegin(), map_.cend(), voxels.begin(),
                   [](const auto &map_element) { return map_element.first; });
    return voxels;
}
```

l00p3 (changes requested):

cpp/kiss_icp/core/VoxelHashMap.cpp
nachovizzo: I never liked this `transform` operations as it confuses me when I read it, we are not transforming anything, we are just extracting data. And 2, it's typically slower, as the `std::vector<Voxel> voxels(map_.size())` will construct `map.size()` voxels calling the default constructor and later on the `std::transfrom` will iterate over all the constructed elements, modifying those in place.

If you have a STL algorithm that sucks the voxels out of the hash map then I take it :nerd_face:

nachovizzo:
> Regarding the visualization, I think the same can be achieved in some way with Polyscope. Probably it will require some massage, but should be faster than the Open3D implementation (I hope). I'll update you as soon as I can.

No rush, thanks for checking!

cpp/kiss_icp/core/VoxelHashMap.cpp
nachovizzo: That's partially not true, my friend; I'm guilty as I was the person promoting this idea at the lab :P . If we check the STL implementation of `std::transform`
```cpp
  /**
   *  @brief Perform an operation on a sequence.
   *  @ingroup mutating_algorithms
   *  @param  __first     An input iterator.
   *  @param  __last      An input iterator.
   *  @param  __result    An output iterator.
   *  @param  __unary_op  A unary operator.
   *  @return   An output iterator equal to @p __result+(__last-__first).
   *
   *  Applies the operator to each element in the input range and assigns
   *  the results to successive elements of the output sequence.
   *  Evaluates @p *(__result+N)=unary_op(*(__first+N)) for each @c N in the
   *  range @p [0,__last-__first).
   *
   *  @p unary_op must not alter its argument.
  */
  template<typename _InputIterator, typename _OutputIterator,
	   typename _UnaryOperation>
    _GLIBCXX20_CONSTEXPR
    _OutputIterator
    transform(_InputIterator __first, _InputIterator __last,
	      _OutputIterator __result, _UnaryOperation __unary_op)
    {
      for (; __first != __last; ++__first, (void)++__result)
	      *__result = __unary_op(*__first);
      return __result;
    }
```

What happens is that sometimes using STL, besides being more precise to express what you want to do, does lead to optimization when compared with our hand-baked shitty code. In this particular case, `const auto &[voxel, voxel_block] : map_` vs. `map_element.first` might be slower, and the STL algorithm forces us to write this in a "better way." But is not doing much more magic in this specific scenario ;)

l00p3: By "optimization" I meant: once compiled probably "reserving and emplacing_back" vs. "pre-constructing and modifing in place" is the same in terms of binary code (but I'm guessing here, no idea how compilation optimizes the two things at the end, probably better from my side to actually check this before guessing and talking). Aware that no magic happens with STL algorithms. In this silly scenario I think is more a matter of taste that practical difference, that was my claim :) And the idea of "trasforming" a voxel_map into a vector of voxel through a `std::transform` was juicy :yum:

nachovizzo: > And the idea of "trasforming" a voxel_map into a vector of voxel through a std::transform was juicy ... 

 start like this, now you tricked my brain

python/kiss_icp/tools/visualizer.py
```diff
@@ -215,15 +234,26 @@ def _update_geometries(self, source, keypoints, target, pose):
...
         else:
             self.target.points = self.o3d.utility.Vector3dVector()
 
+        # VoxelHashMap
+        if self.render_voxel_grid and self.global_view:
+            self.voxel_map.clear()
+            self.voxel_map.voxel_size = self.voxel_size
+            voxels = target_map.get_voxels()
+            for voxel in voxels:
+                self.voxel_map.add_voxel(self.o3d.geometry.Voxel(voxel, LIGHT_GRAY))
```
benemer, line 253: This will require `open3d>v0.18.0`, see https://github.com/isl-org/Open3D/pull/6023, which we have to change in the `pyproject.toml`.

But anyway obsolete if we do the voxel visualization with the new kissualizer :)

nachovizzo: Yes yes, this was a draft proposal with the idea in mind that will be obsolete the moment I open the PR. Thanks for the clarification

benemer:
This is what I think is happening:

The points without voxels that you see are just visualization issues; our internal voxelization and nearest-neighbor search we working properly. Here is why:

Going back to the "wrong" voxelization using `static_cast<int>` without the `std::floor` operation has the following effect assuming a voxel size of 10m: Points with an x coordinate in the interval -10m to 10m will be assigned to an x index of 0 (for that coordinate), whereas the points outside this will be correctly mapped (-20m to -10m is mapped to index -1, 10m to 20m is mapped to index 1 and so on). This leads to a voxel grid that has two different voxel sizes: 20m along the coordinate axes and 10m everywhere else; see Gupta's visualization here: https://github.com/PRBonn/kiss-icp/pull/354#issuecomment-2206113028

How can we explain the points with missing voxels from your visualization? When you ask for our internal map for the occupied voxels, you only return the indices. In the above 1D example, we would return indices -1, 0, and 1. In open3d, you specify that our voxel grid has a fixed voxel size of, for example, 10m so that we will visualize the points in three voxels, each 10m wide, spanning 30m in that direction. This does not reflect that we have larger voxels close to the center, meaning that our voxels actually span 10m+20m+10m=40m in that direction. 

That's why it looks like the points are outside a voxel. Internally, they still are, but the voxel grid is shifted in the visualization.

I tried to visualize what I meant:

image

So internally, we did not have points in completely wrong voxels. But as discussed in https://github.com/PRBonn/kiss-icp/pull/354, the wrong voxelization leads to larger voxels close to the coordinate axes, therefore considering more "volume" when searching for nearest neighbors in that area.

Let me know what you think.

nachovizzo:
> This is what I think is happening:
> 
> The points without voxels that you see are just visualization issues; our internal voxelization and nearest-neighbor search we working properly. Here is why:
> 
> Going back to the "wrong" voxelization using `static_cast<int>` without the `std::floor` operation has the following effect assuming a voxel size of 10m: Points with an x coordinate in the interval -10m to 10m will be assigned to an x index of 0 (for that coordinate), whereas the points outside this will be correctly mapped (-20m to -10m is mapped to index -1, 10m to 20m is mapped to index 1 and so on). This leads to a voxel grid that has two different voxel sizes: 20m along the coordinate axes and 10m everywhere else; see Gupta's visualization here: [#354 (comment)](https://github.com/PRBonn/kiss-icp/pull/354#issuecomment-2206113028)
> 
> How can we explain the points with missing voxels from your visualization? When you ask for our internal map for the occupied voxels, you only return the indices. In the above 1D example, we would return indices -1, 0, and 1. In open3d, you specify that our voxel grid has a fixed voxel size of, for example, 10m so that we will visualize the points in three voxels, each 10m wide, spanning 30m in that direction. This does not reflect that we have larger voxels close to the center, meaning that our voxels actually span 10m+20m+10m=40m in that direction.
> 
> That's why it looks like the points are outside a voxel. Internally, they still are, but the voxel grid is shifted in the visualization.
> 
> I tried to visualize what I meant:
> 
> image
> 
> So internally, we did not have points in completely wrong voxels. But as discussed in #354, the wrong voxelization leads to larger voxels close to the coordinate axes, therefore considering more "volume" when searching for nearest neighbors in that area.
> 
> Let me know what you think.

@benemer, thanks for the excellent explanation; I think it makes sense. Ultimately, it led to more volumetric space to search, which was not the intended behavior. Thanks for clarifying the visualization; I also agree that my reasoning was slightly misled by the fact that I baked the voxels in a completely different module than the internal map, and thus, it looks like this.

nachovizzo:
Closing this in favor of https://github.com/PRBonn/kiss-icp/pull/390 as it got replaced by the new shinny visualizer

PR #388, opened 2024-08-08
Move NN search back to the VoxelHashMap

tizianoGuadagnino:
I simply moved the functions related to the Nearest Neighbor search from ```Registration``` to ```VoxelHashMap```, as this is the data structure that is used to compute the neighbors. This allows other people to use our same strategy of NN search without copy-pasting code from an unnamed namespace. I know that this is something we did a long time ago, but I changed my mind, and this makes much more sense, as now you can use our NN search decoupled from the registration module. Of course, performances are completely unaffected, as I can show here.

comparison

Ps: this change is propedeutic (io non parlo inglese) to a slightly more significant change that will remove a useless copy of points from the voxels during the NN search.

## Related PRs

This PR basically shadows what we did in https://github.com/PRBonn/kiss-icp/pull/290 when we refactored a bit the system to support different number of threads in https://github.com/PRBonn/kiss-icp/pull/252

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -29,20 +29,59 @@
 
 #include "VoxelUtils.hpp"
 
-namespace kiss_icp {
+namespace {
+using kiss_icp::Voxel;
 
-std::vector<Eigen::Vector3d> VoxelHashMap::GetPoints(const std::vector<Voxel> &query_voxels) const {
+std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels = 1) {
+    std::vector<Voxel> voxel_neighborhood;
+    for (int i = voxel.x() - adjacent_voxels; i < voxel.x() + adjacent_voxels + 1; ++i) {
+        for (int j = voxel.y() - adjacent_voxels; j < voxel.y() + adjacent_voxels + 1; ++j) {
+            for (int k = voxel.z() - adjacent_voxels; k < voxel.z() + adjacent_voxels + 1; ++k) {
+                voxel_neighborhood.emplace_back(i, j, k);
+            }
+        }
+    }
+    return voxel_neighborhood;
+}
+std::vector<Eigen::Vector3d> GetPoints(const std::vector<Voxel> &query_voxels,
+                                       const kiss_icp::VoxelHashMap &voxel_map) {
     std::vector<Eigen::Vector3d> points;
-    points.reserve(query_voxels.size() * static_cast<size_t>(max_points_per_voxel_));
+    points.reserve(query_voxels.size() * static_cast<size_t>(voxel_map.max_points_per_voxel_));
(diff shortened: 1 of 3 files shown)


nachovizzo:
I can see us in 3 months going back to the old implementation. I guess we changed this five times already :stuck_out_tongue: haha

Depending on how I wake up, I believe there is no optimal place for this function. I love or hate this change, so let's take it! 

This is also a follow-up on https://github.com/PRBonn/kiss-icp/pull/290#issuecomment-1976227793

I linked the related PR https://github.com/PRBonn/kiss-icp/pull/290 in the description

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -45,6 +58,32 @@ std::vector<Eigen::Vector3d> VoxelHashMap::GetPoints(const std::vector<Voxel> &q
...
 }
 
+}  // namespace
+
+namespace kiss_icp {
+
+std::tuple<Eigen::Vector3d, double> VoxelHashMap::GetClosestNeighbor(
+    const Eigen::Vector3d &point) const {
+    // Convert the point to voxel coordinates
+    const auto &voxel = kiss_icp::PointToVoxel(point, voxel_size_);
```
nachovizzo, line 68:
```suggestion
    const auto &voxel = PointToVoxel(point, voxel_size_);
```

nachovizzo (approved):
Looks good to me!

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -29,14 +29,27 @@
 
 #include "VoxelUtils.hpp"
 
-namespace kiss_icp {
+namespace {
+using kiss_icp::Voxel;
```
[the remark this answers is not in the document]
tizianoGuadagnino: I also agree; I think the alias is mostly for simplifying the understanding, but yeah I also start to don't like it too much.

nachovizzo (approved):

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -29,14 +29,27 @@
...
+        for (int j = voxel.y() - adjacent_voxels; j < voxel.y() + adjacent_voxels + 1; ++j) {
+            for (int k = voxel.z() - adjacent_voxels; k < voxel.z() + adjacent_voxels + 1; ++k) {
+                voxel_neighborhood.emplace_back(i, j, k);
+            }
+        }
+    }
+    return voxel_neighborhood;
+}
+std::vector<Eigen::Vector3d> GetPoints(const std::vector<Voxel> &query_voxels,
+                                       const kiss_icp::VoxelHashMap &voxel_map) {
```
benemer, line 47: This should also be moved to the VoxelHashMap, but we can merge for now. Tiziano and I will do this in a follow-up PR :+1:

benemer (approved):

PR #389, opened 2024-08-08
Remove GetPoints

tizianoGuadagnino:
Hi guys, it is me again. I just removed the GetPoints function, as it is now redundant because we moved the nn search to the ```VoxelHashMap```, which has the same performance but less code.

comparison

To follow up, move the _does the voxel exist_ check from the NN search to the ```AdjacentVoxels``` function. This opens a lot of additional filtering that can be done on the NN search, which should (hopefully) reduce the compute.

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -56,14 +56,14 @@ std::tuple<Eigen::Vector3d, double> VoxelHashMap::GetClosestNeighbor(
     // Find the nearest neighbor
     Eigen::Vector3d closest_neighbor = Eigen::Vector3d::Zero();
     double closest_distance = std::numeric_limits<double>::max();
-    auto closer = [&query](const auto &lhs, const auto &rhs) {
-        return (lhs - query).norm() < (rhs - query).norm();
-    };
-    std::for_each(query_voxels.cbegin(), query_voxels.cend(), [&](const auto &qv) {
-        auto search = map_.find(qv);
+    std::for_each(query_voxels.cbegin(), query_voxels.cend(), [&](const auto &query_voxel) {
+        auto search = map_.find(query_voxel);
         if (search != map_.end()) {
             const auto &points = search.value();
-            const auto &neighbor = *std::min_element(points.cbegin(), points.cend(), closer);
+            const Eigen::Vector3d &neighbor = *std::min_element(
+                points.cbegin(), points.cend(), [&](const auto &lhs, const auto &rhs) {
+                    return (lhs - query).norm() < (rhs - query).norm();
+                });
             double distance = (neighbor - query).norm();
             if (distance < closest_distance) {
                 closest_neighbor = neighbor;
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -43,42 +43,32 @@ std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels = 1
     }
     return voxel_neighborhood;
 }
-std::vector<Eigen::Vector3d> GetPoints(const std::vector<Voxel> &query_voxels,
-                                       const kiss_icp::VoxelHashMap &voxel_map) {
-    std::vector<Eigen::Vector3d> points;
-    points.reserve(query_voxels.size() * static_cast<size_t>(voxel_map.max_points_per_voxel_));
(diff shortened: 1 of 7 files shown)


cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -43,42 +43,32 @@ std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels = 1
...
 }  // namespace
 
 namespace kiss_icp {
 
 std::tuple<Eigen::Vector3d, double> VoxelHashMap::GetClosestNeighbor(
-    const Eigen::Vector3d &point) const {
+    const Eigen::Vector3d &query) const {
     // Convert the point to voxel coordinates
-    const auto &voxel = PointToVoxel(point, voxel_size_);
+    const auto &voxel = PointToVoxel(query, voxel_size_);
```
nachovizzo, line 53: Shouldn't this be called `query_voxel` then? to be conistent

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -43,42 +43,32 @@ std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels = 1
...
-
     // Find the nearest neighbor
     Eigen::Vector3d closest_neighbor = Eigen::Vector3d::Zero();
     double closest_distance = std::numeric_limits<double>::max();
-    std::for_each(neighbors.cbegin(), neighbors.cend(), [&](const auto &neighbor) {
-        double distance = (neighbor - point).norm();
-        if (distance < closest_distance) {
-            closest_neighbor = neighbor;
-            closest_distance = distance;
+    auto closer = [&query](const auto &lhs, const auto &rhs) {
```
nachovizzo, line 59: Why not throw this lambda in the `min_element`  search? It isn't very clear for me to `closer`, and then I had to look where this was used. 

On the other hand, is not very clear what is going on here, too much `auto`, `lhs`, Later on `const auto &neighbor = *std::min_element(points.cbegin(), points.cend(), closer);`.... If we want to go this road I will use more explicit types or try to rewrite this in a clear manner.

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -43,42 +43,32 @@ std::vector<Voxel> GetAdjacentVoxels(const Voxel &voxel, int adjacent_voxels = 1
...
+        return (lhs - query).norm() < (rhs - query).norm();
+    };
+    std::for_each(query_voxels.cbegin(), query_voxels.cend(), [&](const auto &qv) {
```
nachovizzo, line 62: `qv` -> `query_voxel` :)

nachovizzo (commented):
I don't really get why we do this, but I'm also happy to proceed. I left some comments mainly on readability.

I'd also refactor the PR description as we don't remove the `GetPoints` function but absorve its functionality in `VoxelHashMap::GetClosestNeighbor`. Also I didn't understand the comment with the `GetAdjacentVoxels` at all :rofl:

cpp/kiss_icp/core/VoxelHashMap.cpp
benemer: I would also vote for moving it below, then we can also avoid re-computing the distance to the closest neighbor

tizianoGuadagnino:
> I don't really get why we do this, but I'm also happy to proceed. I left some comments mainly on readability.
> 
> I'd also refactor the PR description as we don't remove the `GetPoints` function but absorve its functionality in `VoxelHashMap::GetClosestNeighbor`. Also I didn't understand the comment with the `GetAdjacentVoxels` at all 🤣

Nope bruda, I literally remove it, as it was concatenating the inner points of each voxel into a single ```std::vector```. Now we don't copy this points, but just use a ref from the ```robin_map```.

Thanks for the comments ❤️ I will integrate them now.

About the ```GetAdjacentVoxels```, stay tuned we are working on something with @benemer.

cpp/kiss_icp/core/VoxelHashMap.cpp
tizianoGuadagnino: I have to keep it as ```voxel``` as I want to use ```query_voxel``` into the `for_each`

cpp/kiss_icp/core/VoxelHashMap.cpp
tizianoGuadagnino: Done, tell me if this is better

nachovizzo (approved):

PR #392, opened 2024-08-09
No reallocation

tizianoGuadagnino:
It's a simple fix that lets us gain some performance without changing anything. I quickly benchmarked this, and there was a significant difference. I built it on top of #391 so that it can be merged in sequence. 

https://quick-bench.com/q/kayKiXTVXIT4lK9VmZl0YHnmE9M

❤️ don't hate me guys

### DIFF
--- a/cpp/kiss_icp/core/VoxelHashMap.cpp
+++ b/cpp/kiss_icp/core/VoxelHashMap.cpp
@@ -30,12 +30,16 @@
 #include "VoxelUtils.hpp"
 
 namespace {
+static constexpr int adjacent_voxels = 1;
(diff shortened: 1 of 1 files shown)


benemer (approved):

cpp/kiss_icp/core/VoxelHashMap.cpp
```diff
@@ -30,12 +30,16 @@
 #include "VoxelUtils.hpp"
 
 namespace {
+static constexpr int adjacent_voxels = 1;
```
nachovizzo, line 33: We can drop this as parameters at this point, right :thinking: ?

nachovizzo (approved):

PR #400, opened 2024-10-07
Adds in a class for performing parallel operations with n threads

neoques:
Unclear if I should also limit threads for the tbb::reduce option or if the global limit in the threads is sufficient.


benemer:
Thank you for the PR. Before we have a look at it, can you please check the CI pipeline? It is still failing currently

tizianoGuadagnino:
To be fixed in #412

PR #411, opened 2024-12-04
Improve Runtime by reshaping the DataAssociation

tizianoGuadagnino:
# Motivation

By simply switching the `Correspondences` to a `tbb::concurrent_vector` and performing the `DataAssocation` with a `parallel_for` instead of using a reduction, the pipeline's runtime decreased by 25% when I ran it in a single thread. The fantastic side effect is that the code is also much more readable. 

# Results
## Main
runtime_warehouse_medium_main
## This PR
runtime_warehouse_medium_pr

This is a follow-up PR from #410

### DIFF
--- a/cpp/kiss_icp/core/Registration.cpp
+++ b/cpp/kiss_icp/core/Registration.cpp
@@ -23,8 +23,10 @@
 #include "Registration.hpp"
 
 #include <tbb/blocked_range.h>
+#include <tbb/concurrent_vector.h>
 #include <tbb/global_control.h>
 #include <tbb/info.h>
+#include <tbb/parallel_for.h>
 #include <tbb/parallel_reduce.h>
 #include <tbb/task_arena.h>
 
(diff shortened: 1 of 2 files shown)


cpp/kiss_icp/core/Registration.cpp
```diff
@@ -61,30 +63,17 @@ Correspondences DataAssociation(const std::vector<Eigen::Vector3d> &points,
...
                 if (distance < max_correspondance_distance) {
-                    res.emplace_back(point, closest_neighbor);
+                    correspondences.emplace_back(point, closest_neighbor);
                 }
             });
-            return res;
-        },
-        // 2nd lambda: Parallel reduction
-        [](Correspondences a, const Correspondences &b) -> Correspondences {
-            a.insert(a.end(),                              //
```
tizianoGuadagnino, line 82: This seems to be the problematic operation. It might require more investigation, though. Anyway, I feel that this reduction was inappropriate for the task here.

benemer: What do you mean by problematic? Regarding the runtime?

tizianoGuadagnino: Each insertion might cause a reallocation in the vector `a`.

benemer (approved):

Issue #295, opened 2024-03-18
Modularize Preprocessing Pipeline


nachovizzo:
As exposed in https://github.com/PRBonn/kiss-icp/pull/290 where we made the registration module a class, I think we should go ahead and do the same with the preprocessing module, allowing us for more customization/parametrization of this bit. I'm copying the last part of that PR in here:

From a system point of view, it's nice that we now have this design:
```cpp
// KISS-ICP pipeline modules
Registration registration_;
VoxelHashMap local_map_;
AdaptiveThreshold adaptive_threshold_;
```

Makes me wonder if we shouldn't merge `Deskew` into `Preprocessing` and convert it to another component as well :eyes: ,  similar to how we do it in Python :thinking: , but to be discussed later 

## Items to address

- [ ] Deskew is a preprocessing step, should be there
- [ ] I'd love to move the preprocessing step for Kitti out of our CORE library and just drop it in the Python bindings to avoid people reporting the wrong number in papers
- [ ] Decide if it's worth doing the deskewing with tbb, probably we should benchmark it (I think the answer is yes). If we keep it, we should decide on a clean way to control the number of threads "globally" https://github.com/PRBonn/kiss-icp/pull/252#issuecomment-2003349182

CC @tizianoGuadagnino , @benemer

PR #412, opened 2024-12-05
Preprocessing Pipeline

tizianoGuadagnino:
# Motivation

I finally wanted to tackle #295, as the preprocessing has been an asymmetry in the design for a while, and I would like to have an emotional closure with it. As a side effect, this also allows control of the number of threads for deskewing, as now all preprocessing is performed in a single module (similar to the attempt already made in #400).

# This PR

Introduce the `Preprocessor` module, which is in charge of clipping the scan to the `min`/`max` range and performing the motion compensation. For now, I kept the voxel downsampling outside the preprocessing pipeline, although we could merge it in this module if we like the idea. 

# Results

Of course, I obtained the same numbers in terms of runtime and "accuracy" for the few sequences I tested.


benemer:
Amazing! It looks good to me. Now we also use tbb for deskewing + only deskew the points that are not clipped :)

It's OK to keep the voxel downsampling for now because we also use it in the VoxelHashMap, which is not part of Preprocessing.

benemer (dismissed):

tizianoGuadagnino:
Successive commits represents the fact that me and @benemer realize that the numbers were differing slightly from `main` in terms of accuracy, while the runtime seemed to be affected the most based on single thread vs multi-thread. For these reasons we converged to the current solution which replicates the code present in `main` but wrap it into a function so that we can get both a separate module for the preprocessing and control over the threads used for deskewing, while getting the same numbers for runtime and accuracy.

python/kiss_icp/pipeline.py
```diff
@@ -121,7 +121,7 @@ def _next(self, idx):
             frame, timestamps = dataframe
         except ValueError:
             frame = dataframe
-            timestamps = np.zeros(frame.shape[0])
+            timestamps = np.array([])
```
benemer, line 124: This also fixes a bug: If de-skewing is enabled but no timestamps are provided, the system should not deskew. However, with the default zero timestamps, we still do it and transform the points to the pose at the middle of the scan duration, although the scans might already be de-skewed.

benemer (approved):

Issue #414, opened 2024-12-10
[ROS2] Use relative names for kiss topics


fdila:
Very minor issue: since the various topics that kiss_icp uses begin with "/" they are absolute, and when running kiss_icp with a specified namespace they don't get remapped.

Allowing namespaces would just be a matter of removing the  "/" from the beginning of the name topics here: https://github.com/PRBonn/kiss-icp/blob/0395784225256b2abb03fc591aa11739f96a2fd1/ros/src/OdometryServer.cpp#L115-L120

This modification would be beneficial on a multi-robot setup, since using namespaces is pretty much required to run multiple instances of the same node.

PR #415, opened 2024-12-10
Use relative names for topics

fdila:
Quick PR to address issue #414

Tested locally, if you run kiss_icp without a namespace everything stays as before this PR, if a namespace is specified messages are published on namespace/kiss/

Also updated the .rviz config to match this modifications and also allow to run rviz with a namespace.

### DIFF
--- a/ros/rviz/kiss_icp.rviz
+++ b/ros/rviz/kiss_icp.rviz
@@ -59,7 +59,7 @@ Visualization Manager:
             Filter size: 10
             History Policy: Keep Last
             Reliability Policy: Best Effort
-            Value: /kiss/frame
+            Value: kiss/frame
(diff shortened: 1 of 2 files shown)


benemer:
Looks good to me! Can you please check the CI? I think formatting is failing

fdila:
Weird, I just modified the code from the github editor for the PR (I'm lazy and I have a private fork with lots of edits to suit my needs), seems like one line was already broken. Anyway fixed in the last commit.

benemer:
No, I think the `/` you removed made it possible to fit in one line :)

Now it looks good! Thanks for the fix.

tizianoGuadagnino (approved):

Issue #299, opened 2024-03-24
Deskew model


kulfer80:
Hi, first of all, thank you for your great work!

I have a consideration about de-skew model. I'm not that comfortable in concepts of lie algebra, but I don't totally understand, why do we use parameter `mid_pose_timestamp` ? I have read paper but I didn't find clear explanation. For me the thing that is logical that first point should be just `corrected_frame[0] = motion * frame[0];`, where motion in this case should have been identity transformation matrix. But when using `mid_pose_timestamp` it will yield to different result isn't it ? You are all for sure better experts in this field, but I just wanted to ask this stupid question to understand broader context about this beautiful and well written masterpiece.

https://github.com/PRBonn/kiss-icp/blob/994d2329effaec481b88adf2a0d2aa1a9ec40187/cpp/kiss_icp/core/Deskew.cpp#L44

Thank you for your time and effort.

nachovizzo:
@kulfer80 good question, I have been waiting months until someone asks this.

I'm currently a bit overloaded but I promise I will come back with a detailed explanation. In the meantime, you can rest assured, that 0.5 makes total sense

changh95:
TLDR; Based on my experience, it's critical to use mid_pose_timestamp for effective KISS-ICP operation.

I'm working on a skewed LiDAR dataset, where each LiDAR frame's timestamp was set at the start of its scanning sequence. I used `const auto motion = Sophus::SE3d::exp(timestamps[i] * delta_pose)` to deskew the LiDAR frames, based on the premise that the LiDAR's movement could be approximated by a constant velocity model. This approach seemed reasonable, as it intended to correct distortions more significantly at the end of the scanning sequence than at the beginning.

However, through my experience, I realized a significant pitfall: the constant velocity model is not flawless and introduces errors that get woven into the LiDAR points during the deskewing process. When applied with `const auto motion = Sophus::SE3d::exp(timestamps[i] * delta_pose)`, these errors become proportionally tied to the timestamps, significantly affecting the later points in the sequence. This is particularly problematic for KISS-ICP, which depends on precise point-to-point comparisons, and these errors can escalate, particularly when compounded over several frames.

[...]

PR #418, opened 2025-01-08
Transform the frame at the beginning of scan

tizianoGuadagnino:
Since the beginning of time we decided to deskew the input scans at the middle of the scan duration (*0.5*), the motivation for this was given in #299, and it is an heritage of our initial development based on KITTI (everybody has his dark little secrets). The goal of this PR is to set things (partially) right and deskew at the beginning of the scan (*0.0*), which should be the most common case, at least in the Python setting.


tizianoGuadagnino:
Registration gets slower on some datasets, need investigation.

PR #426, opened 2025-01-22
Remove paris luco dataset and plyfile dependency. Fixes #425

### LINKED ISSUE #425: kiss-icp depends on plyfile (MIT vs GPL v3)
Hi,

During a review of the software licenses used by some of our projects we noted that kiss-icp list `plyfile` as one of their dependencies:

Image

Kiss-ICP is listed as an MIT license but plyfile is GLP v3 as shown in the table below:
```
 kiss-icp            1.1.0       MIT License                                        
 plyfile             1.0.3       GNU General Public License v3 or later (GPLv3+)    
```

This may be seen as conflict between the two licenses (depending on how you interpret GPL, I am not a legal expert). So we were wondering if that is something you guys already went over and decided that the use of plyfile doesn't infringe the kiss-icp license.

nachovizzo:
As discussed in https://github.com/PRBonn/kiss-icp/issues/425, we removed the Paris Luco datasets, as it's the only dataset that pulls the `plyfile` library dependency, which might bring licensing issues


tizianoGuadagnino:
@nachovizzo, let's goooo <3

tizianoGuadagnino (approved):

benemer (approved):

Issue #176, opened 2023-06-22
Add support to read timestamps from pcd/ply files


nachovizzo:
If deskewing is required, the current generic dataloader is not loading (and thus, discarding) the timestamp information. Should be easy to do it. I tried with PlyData but it's really slow! 

snippet that works (but too slow)
```py
cloud = PlyData.read(file_path)
data = cloud["vertex"].data
points = np.vstack([data[name] for name in ["x", "y", "z"]]).T
timestamps = np.asarray(data["t"])
timestamps = timestamps / timestamps.max()
return points.astype(np.float64), timestamps
```

The old API from open3d also does not support this, maybe migrating to the tensor API helps (?)

iegorval:
Is this still not implemented / not having anyone working on it?

PR #428, opened 2025-01-24
Read Timestamps from PLY

tizianoGuadagnino:
**Motivation**

Finally, I wanted to handle #176, which has been pending since the beginning of time; this PR allows us to read timestamps from a PLY using the open3d tensor API. I use the same predefined fields that we use for ROS message stamping, hoping that this covers 90% of the scenarios. This is now directly incorporated into the generic dataloader for transparent usage from the user side. 

@benemer and I have already tested this on car data recorded here in the lab, and it looks good.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   pre-commit:
     name: Pre-commit checks
-    runs-on: ubuntu-24.04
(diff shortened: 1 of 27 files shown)


benemer (approved):
Looks good to me. I am not super happy with the `ReadOpen3d` class inside a class, but I can't come up with a better idea now. Let's move forward :)

PR #433, opened 2025-01-26
Move the normalization of the timestamps in preprocessing

tizianoGuadagnino:
### Motivation

The user or a dataloader should not handle timestamp normalization; this should be part of our pipeline. My opinion is that our system should process the data as raw as possible without requiring external manipulations. Furthermore, with this change, we can remove potential sources of errors as the user might not necessarily be aware that timestamps need to be in the range $\[0, 1\]$.

### This PR

Introduction of a timestamp normalization step within the `Preprocessor`. This computation is performed just if `deskew=True` and `timestamps` are not empty. As a result, timestamps normalization is not required anymore in the `generic` and ROS-based dataloaders (`mcap` and `rosbag` through the `tools/point_cloud2.py` file).

### DIFF
--- a/python/kiss_icp/datasets/mcap.py
+++ b/python/kiss_icp/datasets/mcap.py
@@ -60,7 +60,7 @@ class McapDataloader:
     def __getitem__(self, idx):
         msg = next(self.msgs).ros_msg
         self.timestamps.append(self.stamp_to_sec(msg.header.stamp))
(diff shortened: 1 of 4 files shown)


python/kiss_icp/datasets/mcap.py
benemer, line 1: good catch!

PR #434, opened 2025-01-27
Remove max_range and deskew from CLI

benemer:
After deprecating the `--deskew` option from the CLI in https://github.com/PRBonn/kiss-icp/pull/432, I think it's time to remove the `--max_range` option too. I don't know how widely this was used (at least I didn't), but it's much cleaner without it because now the CLI options are pipeline-specific and do not overwrite the configuration of KISS-ICP.

However, I would like to keep this as a draft (for now) and see if anyone has objections. If so, please comment here :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ kiss_icp_pipeline --help
 <details>
 <summary>This should print the following help message:</summary>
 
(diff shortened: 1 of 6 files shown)


benemer:
Amazing, ready for final review @saurabh1002

Issue #440, opened 2025-02-18
Bug in "generic.py" .bin Reader Lambda Returns Tuple Instead of Callable


Marvin1412:
In the kiss_icp/datasets/generic.py file, within the _get_point_cloud_reader() function, the KITTI reader for .bin files is defined as follows:

```
if self.file_extension == "bin":
    print("[WARNING] Reading .bin files, the only format supported is the KITTI format")
    return lambda file: np.fromfile(file, dtype=np.float32).reshape((-1, 4))[:, :3], np.array([])
```

Due to the comma outside the lambda, this returns a tuple (lambda, np.array([])) instead of a callable that returns a tuple (points, timestamps). Consequently, when _read_point_cloud(file) is called later, a TypeError: 'tuple' object is not callable is raised.

Proposed Fix:
Wrap the return tuple inside the lambda so that it returns the tuple when called:

```
if self.file_extension == "bin":
    print("[WARNING] Reading .bin files, the only format supported is the KITTI format")
    return lambda file: (np.fromfile(file, dtype=np.float32).reshape((-1, 4))[:, :3], np.array([]))
```

This change ensures that _read_point_cloud is a callable function that returns the expected tuple of (points, timestamps).

tizianoGuadagnino:
Please check #441

PR #441, opened 2025-02-23
Use callable for generic dataloader

tizianoGuadagnino:
### Motivation

This PR handles #440, and proposes to use a callable object for each possible filetype/library in the `generic` dataloader. This gives a pattern to the user and avoids silly syntax mistakes that a lambda can introduce. 

@marvin1412 can you verify that this also solves your issue?


Marvin1412:
It solves my problem, thanks for including it.

benemer (approved):

PR #442, opened 2025-02-27
Fix open3d not raising an exception on invalid files when guessing dataset format

Adraub:
If the file is unreadable the code doesn't continue to read cloud with other available libraries

### DIFF
--- a/python/kiss_icp/datasets/generic.py
+++ b/python/kiss_icp/datasets/generic.py
@@ -86,6 +86,10 @@ class GenericDataset:
             import open3d as o3d
 
             try_pcd = o3d.t.io.read_point_cloud(first_scan_file)
+            if try_pcd.is_empty():
+                #open3d binding does not raise an exception if file is unreadable or extension is not supported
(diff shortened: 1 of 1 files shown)


benemer (approved):
Good catch!

There might be still cases in which the data is corrupted but with a valid header that still passes this check, but I can not think of a better check right now.

python/kiss_icp/datasets/generic.py
```diff
@@ -86,6 +86,10 @@ def _get_point_cloud_reader(self):
             import open3d as o3d
 
             try_pcd = o3d.t.io.read_point_cloud(first_scan_file)
+            if try_pcd.is_empty():
+                # open3d binding does not raise an exception if file is unreadable or extension is not supported
+                raise Exception("Cloud is empty")
```
tizianoGuadagnino, line 91:
```suggestion
                raise Exception("Generic Dataloader| Open3d PointCloud file is empty")
```

tizianoGuadagnino (approved):

PR #445, opened 2025-03-13
Gupta/fix serialized dataloaders

saurabh1002:
#### This PR adds a reset method for serialized dataloaders, namely **rosbag** and **mcap**
- This is useful when we need to go through the dataset again in the same pipeline
- For example during a global mapping run at the end of the pipeline
- The other dataloaders are indexed and therefore allow directly using index `0` for starting over, but rosbags cannot do that

### DIFF
--- a/python/kiss_icp/datasets/mcap.py
+++ b/python/kiss_icp/datasets/mcap.py
@@ -76,6 +76,7 @@ class McapDataloader:
         )
 
     def reset(self):
+        self.timestamps = []
         self.bag = self.make_reader(open(self.mcap_file, "rb"))
         self.msgs = self.read_ros2_messages(self.mcap_file, topics=[self.topic])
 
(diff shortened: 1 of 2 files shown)


python/kiss_icp/datasets/mcap.py
```diff
@@ -42,15 +42,18 @@ def __init__(self, data_dir: str, topic: str, *_, **__):
...
+        self.read_ros2_messages = read_ros2_messages
+        self.read_point_cloud = read_point_cloud
+
+        self.bag = self.make_reader(open(self.mcap_file, "rb"))
         self.summary = self.bag.get_summary()
         self.topic = self.check_topic(topic)
         self.n_scans = self._get_n_scans()
-        self.msgs = read_ros2_messages(mcap_file, topics=[self.topic])
+        self.msgs = self.read_ros2_messages(self.mcap_file, topics=[self.topic])
         self.timestamps = []
```
benemer, line 56: Same here, also reset this please

benemer (approved):

tizianoGuadagnino (approved):

PR #450, opened 2025-03-25
Fix tf frame calculation when base_frame is given

ksuszka:
There seems to be an error in TF frame calculation when base_frame is given.

Running current code with base-lidar frame transformation:
```
ros2 run tf2_ros static_transform_publisher --x 0 --y 0 --z 10 --roll 0 --pitch 0 --yaw 0 --frame-id base_link --child-frame-id os_lidar
```
gives following result:

image

After proposed change it gives following result:
image

### DIFF
--- a/ros/src/OdometryServer.cpp
+++ b/ros/src/OdometryServer.cpp
@@ -156,7 +156,7 @@ void OdometryServer::PublishOdometry(const Sophus::SE3d &kiss_pose,
     const auto pose = [&]() -> Sophus::SE3d {
         if (egocentric_estimation) return kiss_pose;
         const Sophus::SE3d cloud2base = LookupTransform(base_frame_, cloud_frame_id, tf2_buffer_);
-        return cloud2base * kiss_pose * cloud2base.inverse();
+        return kiss_pose * cloud2base.inverse();
     }();
 
     // Broadcast the tf ---


benemer:
I think this is a different issue. What you changed here should be wrong, because of the following:

`kiss_pose` transforms from the `os_lidar` to the `lidar_odom_frame` which is the initial pose of the `os_lidar` frame. In case the `os_lidar` is not the `base_link`, we want to express the pose of `base_link` w.r.t to the initial pose of the `base_link`. We need to do two transformations since we change both reference frames (left and right).

What you observe here is another problem: You artificially put the sensor 10m along z in the air. That is why the frame is now shifted. The reason the map is still at the base frame is that we use the wrong `frame_id` in [this](https://github.com/PRBonn/kiss-icp/blob/1e882314e1cdc2816b9ca8eff0ba49e2f16574ff/ros/src/OdometryServer.cpp#L204) line:

```cpp
    auto local_map_header = header;
    local_map_header.frame_id = lidar_odom_frame_;
    map_publisher_->publish(std::move(EigenToPointCloud2(kiss_map, local_map_header)));
 ```

Here, `local_map_header.frame_id` should not be the frame in which we output the odometry (which is the initial 
 pose of `base_link` in case you provide a `os_lidar` to `base_link` transform) but the first frame of the sensor, in which the odometry is estimated before being transformed w.r.t. the desired frame.

We probably need to store the frame in which the odometry is estimated, which can differ from the `lidar_odom_frame` in which the odometry is published.

**Comment by ksuszka** [comment]
> we want to express the pose of `base_link` w.r.t to the initial pose of the `base_link`. 

Is there any reason for that? Does the initial pose of base_link has any significant meaning? With the proposed change the reference frame is lidar_odom_frame, for both the lidar frame and base_link frame. IMHO it is mathematically correct. Odom frame is just an arbitrary chosen frame, in case of this repository it is directly relevant to lidar frame as it is used to track movement between consecutive lidar readings. All other frames can be derived from it. With this approach you can even have lidar on moving arm (the lidar<->base_link can change dynamically) and AFAIK calculations are still correct.

There is one additional thing that needs to be fixed - LookupTransform function shoud not return identity tf if no lidar<->base_link tf is available as it is factually incorrect (I [prepared new code](https://github.com/ksuszka/kiss-icp/tree/do-not-publish-wrong-odometry), but I didn't created PR yet, as it is based on the change from this PR)

**Comment by benemer** [comment]
> Is there any reason for that? Does the initial pose of base_link has any significant meaning?

See REP 105 [here](https://www.ros.org/reps/rep-0105.html). It is mainly relevant for multi-sensor setups, where, for example, wheel odometry is published as the transformation between `base_link` and `odom`. Here, `odom` is usually fixed at the first pose of the base. You can also have a look at the PR https://github.com/PRBonn/kiss-icp/pull/232 and this issue https://github.com/PRBonn/kiss-icp/issues/174 to see why we changed it.

I agree it is mathematically correct, but I am not sure if that is what we want. It is a bit strange to me to express the pose of the base w.r.t. the initial pose of the sensor. For example, to now fuse LiDAR odometry with wheel odometry, we would need to apply an additional transformation to express the LiDAR odometry w.r.t. `odom`. In my view, it is easier if `odom` and `odom_lidar` refer to the **physically same** odometric frame, the one is used for wheel odometry and the other one for LiDAR odometry.

> With this approach you can even have lidar on moving arm (the lidar<->base_link can change dynamically) and AFAIK calculations are still correct.

This is true, this is something we currently can not handle as discussed in https://github.com/PRBonn/kiss-icp/issues/243 and https://github.com/PRBonn/kiss-icp/issues/243.

However, I am open to discuss this. At least, this PR solves the issue of the map frame (because now the `lidar_odom_frame` is always the frame in which the map lives and in which we estimate the odometry) and the issue about a non-static `cloud2base`. I summon @nachovizzo and @tizianoGuadagnino for their opinions :)

**Comment by ksuszka** [comment]
Just to add some fuel to the discussion ;-)

> In my view, it is easier if `odom` and `odom_lidar` refer to the **physically same** odometric frame

I don't think it is feasible in a long-term as per [REP 105](https://www.ros.org/reps/rep-0105.html#id15). The odom frame is bound to drift due to inaccuracies of the odometry source used to determine pose in that frame. So if you have two odometry sources and you track each one in a separate odom frame then it doesn't matter if at some point in time (i.e. at the start) they had the same numerical pose in both frames, because after some time robot poses relative to each frame origin will differ, but you know that they describe the same robot, so it means that origins of both odom frames drifted from each other. And IMHO this is by REP 105 design and its explained as "The pose of a mobile platform in the odom frame can drift over time, without any bounds. This drift makes the odom frame useless as a long-term global reference.".

If you do want to have single odom frame build from two odometry sources then you need to implement something like EKF which takes as input relative movment data from both sources and incorporates them into single pose prediction in some arbitrary choosen odom frame.

In other words, as by REP 105 design, because in a long-term the origin of odom frame drifts relative to the current robot pose the initial value of odom origin doesn't matter.

At least this is my intepretation of what is described in REP 105.

**Comment by benemer** [comment]
The `odom` frame does not drift, it is world-fixed. The pose of the robot drifts with respect to `odom`.

Of course, if you only care about relative odometry, the reference frame does not matter. However, it still feels wrong to me to publish the base's pose with respect to the LiDAR's first pose. But that might also be a matter of taste.

**Comment by ksuszka** [comment]
> The `odom` frame does not drift, it is world-fixed. The pose of the robot drifts with respect to `odom`.

That depends on what you understand by "world". It does drift w.r.t map frame/physical world and it is descibed [here](https://www.ros.org/reps/rep-0105.html#frame-authorities): 
> The transform from odom to base_link is computed and broadcast by one of the odometry sources.
>
>The transform from map to base_link is computed by a localization component. However, the localization component does not broadcast the transform from map to base_link. Instead, it first receives the transform from odom to base_link, and uses this information to broadcast the transform from map to odom.

So the origin of the odom frame drifts in time in the physical world. It has to drift because REP 105 imposes following contraints:
- map<->base_link is derived from absolute position measurement and can jump
- odom<->base_link is derived from relative position measurement and cannot jump
- so if the TF tree is map->odom->base_link and a new absolute measurement is received which causes absolute position of the base_link to jump in the map frame then this jump can only be applied to map->odom part of the tree because odom->base_link part must be continuous. So the odom frame origin in the map frame (representing the physical world) has to jump.

**Comment by ksuszka** [comment]
> Can you give me an example in which it is relevant that the LiDAR odometry is published as a transformation between the robot's base and the initial pose of the LiDAR scanner instead of just the initial pose of the base? Maybe that will help me understand your motivation behind this change.

So to be clear. As I understand the current situation:
- The current code is wrong when base_frame != lidar_frame, which is demonstrated by this exaggerated example I provided in the first post.
- This PR tries to correct this code and make it mathematically correct with minimal changes, but I understand that it may be counter-intuitive, depending on the personal view on the odom frame expectations.
- Fixing current code by introducing separate odom frame with its origin in the initial pose of the base frame works for me too, although IMHO it is not necessary hence the whole "odom drift" discussion

I originally stumbled upon this issue when I tried to replace the visual odometry component from [mrg_slam](https://github.com/aserbremen/mrg_slam) package with kiss_icp. With the proposed changes I got very good results on preparing PCD map for a 0.5km2 of a terrain in my area.

**Comment by benemer** [comment]
If it does not matter or has no advantage, I see no reason to have two different odometric frames for estimating the `base_link` odometry. In that case, I can not visualize the locations of the odometry estimates of different sources (for example, LiDAR and wheel) because they are published with different odometric frames.

Let's wait for some more opinions on that.

**Comment by benemer** [comment]
@tizianoGuadagnino can you have a look at this?

**Comment by adityapatil-peppermint** [comment]
@ksuszka If you are visualizing `/kiss/local_map` (Grey PCL) and `/kiss/frame` (Cyan PCL), the problem might be related to transformation in visualization of `/kiss/local_map`. `/kiss/local_map` is computed in lidar frame but during publishment `lidar_odom_frame` which is `odom` in your case I assume.

**Comment by adityapatil-peppermint** [comment]
We need to transform the kissp_map to robot base frame before publishment
Updated `PublishClouds` method would look something like this.

```
void OdometryServer::PublishClouds(const std::vector<Eigen::Vector3d> &frame,
                                   const std::vector<Eigen::Vector3d> &keypoints,
                                   const std_msgs::msg::Header &header) {
    const auto kiss_map = kiss_icp_->LocalMap();

    frame_publisher_->publish(std::move(EigenToPointCloud2(frame, header)));
    kpoints_publisher_->publish(std::move(EigenToPointCloud2(keypoints, header)));
    auto local_map_header = header;
    local_map_header.frame_id = lidar_odom_frame_;
    
    // We need to transform the kissp_map to robot base frame before publishment
    Sophus::SE3d T = LookupTransform(base_frame_, header.frame_id, tf2_buffer_);
    map_publisher_->publish(std::move(EigenToPointCloud2(kiss_map, T, local_map_header)));
}
```

PR #451, opened 2025-03-27
Do not publish odometry if base<->lidar tf is missing

ksuszka:
In current code, if base_frame is different from cloud_frame_id, but proper TF between base_frame and cloud_frame_id is not yet received, then odometry is published with an assumption that base_frame is equal to cloud_frame_id, which is usually wrong.
After this change if base_frame is different from cloud_frame_id, the odometry/TF won't be published until base_frame<->cloud_frame_id transformation is known.

Note: I left pose transformation as in the original code, but IMHO it is wrong and should be fixed by https://github.com/PRBonn/kiss-icp/pull/450 or by some other way.

### DIFF
--- a/ros/src/OdometryServer.cpp
+++ b/ros/src/OdometryServer.cpp
@@ -22,6 +22,7 @@
 // SOFTWARE.
 #include <Eigen/Core>
 #include <memory>
+#include <optional>
 #include <sophus/se3.hpp>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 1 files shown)


benemer:
Thank you for the PR, and sorry for the late reply!

Shouldn't it be the responsibility of the user to provide that tf, if they specify a different `base_frame`? I'm not sure what's better: assuming identity and at least getting some estimate, or not publishing anything.

ksuszka:
Yes, it is the reponsibility of the user to provide this TF. This PR fixes the issue when user doesn't provide the required TF on time and kiss_icp returns the wrong answer.

As ROS is a distributed system, there is no guarantee that TF is received before cloud data and the code shoudn't depend on it.

The current implementation makes the output of kiss_icp unreliable, as the client cannot easily determine if the TF received from kiss_icp is the correct one, or "the default" one which is most likely wrong.

benemer:
I am still not sure what the better option is: Either not publishing anything if the TF is not available, like in this PR, or assuming identity to still output odometry. For example, some users might use `base_frame` for just renaming the moving frame, in this case, the identity is correct.

ksuszka:
It may be a matter of point of view. I will just state mine.

If I have to choose between:
-  a function which always return a value, but sometimes it is wrong and there is no way to determine if the returned value is right or wrong, and
- a function which returns only a correct value or no value at all

I will always choose the latter.

And even if we extend it to:
-  a function which always return a value, but sometimes it is wrong and there is no way to determine if the returned value is right or wrong, but there may be cases for some subset of input data that the function is always right and
- a function which returns only a correct value or no value at all, and returns no value at all for those cases of a subset of input data that it would return the correct answer,

I will still choose the latter.

For me reliability is much more important than availability.

> For example, some users might use base_frame for just renaming the moving frame, in this case, the identity is correct.

Even a broken clock is right twice a day ;)

PR #464, opened 2025-05-07
Add parameters for correcting published odometry timestamps

ksuszka:
Note: Please correct me if this PR is based on wrong assumptions.

The current version of kiss_icp algorithm is deskewing the input point cloud to the youngest (the latest) point. So, the estimated pose is much closer to the pose the robot had at the moment of time the last point was acquired than the moment of time the first point was acquired. However, the current implementation of ROS2 wrapper uses the input point cloud timestamp for published odometry. But some lidars use the start of the acquisition as the timestamp for the point cloud message. So published odometry has effectively timestamps offseted by one frame.

This PR introduces two parameters for ROS2 wrapper:
- timestamp_correction_offset - the absolute timestamp offset to be added to published messages
- timestamp_correction_delta_ratio - the ratio of the delta time between two latest frames to be added to published messages

So, setting timestamp_correction_delta_ratio=1.0 should solve the issue of timestamp "offset by one frame" for point clouds which are stamped at the start.

Note: the best solution would be to use timestamp from the last point, however AFAIK there is no standard regarding the absolute value of that timestamp fiels in the point cloud. The current deskewing code handles this lack of the standard nicely by using it only for relative interpolation so the meaning of actual absolute values doesn't really matter.


ksuszka:
> I don't fully get the purpose of the two variables. Can you explain why a ratio not equal to 1 can be beneficial? And how about the offset?
> One problem I see is that the stamp will be messed up in case of a frame drop.

So the `timestamp_correction_offset` can be used when you know lidar's frequency. The benefit of using it is that you are immune to frame drops as you noticed. But the downside is that you need to change if you happen to change lidar's frequency (and in our case we have some recordings with 10Hz and other with 20Hz)

For the `timestamp_correction_delta_ratio` I can see three reasonable values:
- 0.0 - if your point clouds are timestamped at the end
- 1.0 - if your point clouds are timestamped at the beginning
- 0.5 - if your point clouds are timestamped at the middle of the acqusition (I'm not sure how common it is in practice, but I think I read somewhere that some public dataset use this scheme)

I had some thoughts about the frame drop issue when using `timestamp_correction_delta_ratio` parameter and I considered for a moment to add a buffer with last N deltas and select min/avg delta, or use something like [exponential moving average](https://en.wikipedia.org/wiki/Exponential_smoothing), but I decided to keep it simple. However if you think that adding some mechanism to handle dropped frames is required I can try to add it.

PR #467, opened 2025-05-19
Allow to set ROS parameters from a yaml file

leroycorentin:
Currently, the parameters exposed by the ROS2 node are not handled as proper parameters. The only way to change their value is to modify the launch file and recompile the package.
I made some changes to allow to set all the parameters in a yaml configuration file for more flexibility.


benemer:
Thank you for the feedback. When playing around with your proposed solution, I came across a few things:

- Even if ROS-specific parameters, such as "lidar_odom_frame", are now passed via the CLI, they are being overwritten by the `config.yaml`. If one changes the order of the parameters being passed to the Node in the launch file, the parameters from the `config.yaml` are ignored. Both cases are not ideal, and I would propose separating ROS arguments (only set via CLI) and KISS core parameters (set via a YAML file)
- The only way to change the parameters now is to modify them in the installed configuration file. How about allowing a path to a YAML file via the CLI? We can still use the installed one as the default if the user does not specify a custom config file.

benemer (approved):

ros/config/config.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+    ####    Core KISS-ICP parameters    ####
+
+    data:
+      deskew: True
+      max_range: 100.0
+      min_range: 0.0
+
+    mapping:
+      voxel_size: 1.0 # <- optional
```
saurabh1002, line 24: This fixes the voxel size to 1.0 and does not change it automatically with a change in max_range, which is the default behaviour in KISS-ICP (`voxel_size = max_range / 100`)

image
image

The `voxel_size` parameter either needs to be commented out so that the default calculated from `max_range` is used, or we need to have to separate config files (`basic.yaml` and `advanced.yaml`) like in the python config files for KISS-ICP.

saurabh1002: image

I propose something like this to be the easiest solution. What do you think @leroycorentin @benemer ?

saurabh1002 (changes requested):
Looks good to me in general. Just one fix needed with the `voxel_size` parameter.

ros/config/config.yaml
benemer: I would say let's go for the optional one, that's how we do it on the Python side as well

saurabh1002 (approved):

saurabh1002:
Done, thanks @leroycorentin!!

PR #471, opened 2025-06-07
Update visualizer callback to fit for PolyScope API changes

Athameral:
In `PolyScope v2.4.0`, structure `PickResult` has changed (possibly extended, but I failed to find a historical documentation), thus we cannot treat it as a tuple and unpack it, or there will be a TypeError.

See [this doc](https://polyscope.run/py/basics/interactive_UIs_and_animation/#picking-selection-and-querying-the-scene) for details.

Maybe some version ascertainment is required.


benemer:
Thanks! We already have an open PR about this: https://github.com/PRBonn/kiss-icp/pull/469 and will fix it there.

PR #472, opened 2025-06-10
pass by reference the pointclouds

saurabh1002:
The `OdometryServer::PublishClouds` function was passing the point clouds by value. This PR addresses the issue by passing it as a reference.


benemer (approved):
Good catch!

Issue #478, opened 2025-07-24
how to reset the map/odometry through service??


aishwarya-peppermint:
how to reset the map/odometry through service?? Using ROS2

benemer:
We currently do not provide that functionality, however, it should be straightforward to add it. Unfortunately, I do not have much time at the moment. Feel free to do it yourself and open a PR. I also added the labels such that others feel invited to do so :)

PR #479, opened 2025-08-15
feat: add ROS service to reset odometry

qza36:
### Context
This pull request implements a feature to reset the map and odometry via a ROS service. This functionality was originally discussed and requested in the closed issue **#478**.

This work was undertaken following the maintainer's encouraging comment in that thread ([link to comment](https://github.com/PRBonn/kiss-icp/issues/478#issuecomment-3125864264)):
> We currently do not provide that functionality, however, it should be straightforward to add it. Unfortunately, I do not have much time at the moment. Feel free to do it yourself and open a PR. I also added the labels such that others feel invited to do so :)

### Changes Made
- Added a ROS 2 service named `/kiss/reset` (of type `std_srvs/srv/Empty`).
- When the service is called, the following actions are performed:
    - The odometry pose estimate (`last_pose_`) is reset to the origin (identity transformation).
    - The internal local map (`local_map_`) is completely cleared of all points.
    - The adaptive threshold mechanism (`adaptive_threshold_`) is re-initialized to its default configured state.

### How to Test
1. Launch the `odometry.launch.py`.
2. Drive the robot around or play a rosbag to ensure odometry is being published.
3. In a separate terminal, call the service: `ros2 service call /kiss/reset std_srvs/srv/Empty`.
4. The odometry should reset to the origin. This can be verified by observing the `/tf` tree (e.g., `odom_lidar` -> `base_link` transform becomes identity) or the odometry topic messages.


benemer:
Thanks for the PR! Can you have a brief look at the CI? I think the formatting breaks our style check. Thanks!

qza36:
Hi, thanks for reviewing! My apologies for the style error. I've just pushed a fix for the formatting. Let me know if there is anything else!

benemer (approved):
Looks good to me!

The only thing I could suggest is to use

```cpp
void OdometryServer::ResetService(
    [[maybe_unused]] const std::shared_ptr<std_srvs::srv::Empty::Request> request,
    [[maybe_unused]] std::shared_ptr<std_srvs::srv::Empty::Response> response) {
```

instead of 

```cpp
void OdometryServer::ResetService(const std::shared_ptr<std_srvs::srv::Empty::Request> request,
                                  std::shared_ptr<std_srvs::srv::Empty::Response> response) {
    (void)request;   // Suppress unused parameter warning
    (void)response;  // Suppress unused parameter warning
```
    
which is available from C++17. But that's more a cosmetic thing and I am fine with both.

I will wait for @saurabh1002 for his feedback and then we can merge :)

ros/src/OdometryServer.cpp
```diff
@@ -223,11 +227,22 @@ void OdometryServer::PublishClouds(const std::vector<Eigen::Vector3d> &frame,
...
     frame_publisher_->publish(std::move(EigenToPointCloud2(frame, header)));
     kpoints_publisher_->publish(std::move(EigenToPointCloud2(keypoints, header)));
-
     auto local_map_header = header;
     local_map_header.frame_id = lidar_odom_frame_;
     map_publisher_->publish(std::move(EigenToPointCloud2(kiss_map, local_map_header)));
 }
+void OdometryServer::ResetService(const std::shared_ptr<std_srvs::srv::Empty::Request> request,
+                                  std::shared_ptr<std_srvs::srv::Empty::Response> response) {
+    (void)request;   // Suppress unused parameter warning
```
mehermvr, line 236: i think `[[maybe_unused]]` should work on the parameters. or maybe /* request */ should also work i think. but this is a minor style thing.

mehermvr: ah i just saw then benedikt also made the same comment before. then its fine.

mehermvr (approved):
lgtm as well

benemer:
Thanks, guys! @qza36, I think you have to pull the latest changes from main, such that the CI also runs the tests on kilted and jazzy. We need those to pass to merge. Thanks :)

PR #490, opened 2026-02-26
feat [ROS]: pub output as pose with cov msg type

hellovuong:
Pub result as PoseWithCovarianceStamped

### DIFF
--- a/ros/src/OdometryServer.cpp
+++ b/ros/src/OdometryServer.cpp
@@ -90,6 +90,8 @@ OdometryServer::OdometryServer(const rclcpp::NodeOptions &options)
     // Initialize publishers
     rclcpp::QoS qos((rclcpp::SystemDefaultsQoS().keep_last(1).durability_volatile()));
     odom_publisher_ = create_publisher<nav_msgs::msg::Odometry>("kiss/odometry", qos);
+    pose_publisher_ =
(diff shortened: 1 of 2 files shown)


mehermvr:
an odom message type includes the pose with covariance type anyways, so i don't see how this is relevant

hellovuong:
> an odom message type includes the pose with covariance type anyways, so i don't see how this is relevant

@mehermvr Thank for the review. One thing about the Odometry msg, it contains twist values. It is currently empty in Kiss-ICP so Pose with covariance is a more suitable and more specific msg type for the output of Kiss-ICP.

An example for this msg type is using Kiss-ICP odom output as a constraint in [fuse](https://github.com/locusrobotics/fuse) pkg for sensor fusion optimisation-based. If the input is Odometry msg. They will interpret it as Pose and Twist w/ cov. The optimisation will failed miserably in this case because the twist is always zero. 

Just an example of better using message type in ROS1/2 since a lot of package expected data based on msg type.

mehermvr:
While thats a fair point about kiss icp not publishing the twist right now, the pose with covariance message can only convey one of the two frames involved in the odom<-lidar chain (in the header). The odom message type has an additional field for the child frame. its a tradeoff, and having twist be 0 is the better one in my perspective (which can also be "fixed" and published from the constant velocity guess, but thats a separate feature/pr).
I leave it to the others to comment further on this.

Issue #491, opened 2026-03-17
Consider rosbags.typestore definitions


ExpectedArgument:
In rosbags v0.11.0, `rosbags.typesys.types` is deprecated, removed and replaced with `rosbags.typesys.store`. So, updating the repo may improve compatibility.

https://gitlab.com/ternaris/rosbags/-/commit/de1d1663724a296d354855f6cccdb0afbb061bff

mawerid:
I have the same issue.

Enabling rosbags 0.11 means enabling the use of KISS-ICP with ROS 2 Jazzy, for example.

mehermvr:
Please make a PR with the suggested changes. Would be happy to review it

mawerid:
@mehermvr 

There are no errors in the current main branch, only in the 1.2.3 tag.

So, when will the new release be?

mehermvr:
Good question. I'll bring it up with the other maintainers, so hopefully soon

mehermvr:
@mawerid the new release is live! so hopefully that also solves the original issue @ExpectedArgument

PR #492, opened 2026-04-25
Bump version to 1.3.0

mehermvr:
It's been a while since the last package release. The current version of kiss_icp (python) especially doesnt have the version pinned for rosbags, which can be inconvenient to fix for users. See #488 and #491.

The changes since last release are significant enough with new features (see #467, #479) to warrant a minor version bump.


saurabh1002 (approved):
Lgtm

PR #493, opened 2026-04-25
Remove ament_target_dependencies and replace with target_link_libraries

mehermvr:
`ament_target_dependencies` was deprecated in ros kilted (or jazzy, not sure). And now seemingly removed in rolling (#492 fails ci on rolling). This should just fix that. Needs to be merged in before #492 can be.


saurabh1002 (approved):
Lgtm

PR #494, opened 2026-04-27
Optimize by removing extra loop and using emplace

saurabh1002:
This pull request mainly refactors `Update` function in the `VoxelHashMap` to improve code efficiency in how voxel maps are updated.

* Refactored `VoxelHashMap::Update` to transform and insert points into the voxel map in-place, avoiding creation of an intermediate transformed points vector and using `emplace` for more efficient map insertion.
* Changed `VoxelHashMap::AddPoints` and `VoxelDownsample` to use `emplace` instead of `insert` for more efficient and concise map population.
* Simplified return statements in `VoxelHashMap::GetClosestNeighbor` and `BuildLinearSystem`


saurabh1002:
Nevermind, all of this is useless. Only helped me to learn about possible SIMD optimizations.

PR #500, opened 2026-05-09
Fix division by zero in deskew when all per-point timestamps are equal

gistrec:
## Summary

- Guard the C++ deskew path (`cpp/kiss_icp/core/Preprocessing.cpp`) so `(t - min) / (max - min)` cannot divide by a zero — or non-finite — time range.
- Apply the same guard to the ROS helper `NormalizeTimestamps` (`ros/src/Utils.hpp`), mapping the degenerate case to `1.0`, i.e. the end of the sweep, which deskews to the identity.
- Add `python/tests/test_preprocess.py` covering the degenerate cases plus a positive control.

## Why

Some LiDAR drivers, and plenty of downstream filters, emit point clouds where every point carries the same timestamp. `Preprocess` then computes `0 / 0` for every stamp.

**The symptom is a hard crash, not silent degradation.** `Sophus::SE3d::exp` on a NaN tangent trips `SOPHUS_ENSURE`, whose default handler calls `std::abort()`. That check is not gated on `NDEBUG` and kiss-icp never defines `SOPHUS_DISABLE_ENSURES`, so a release build, a pip wheel and the ROS node all abort with exit 134:

```
[sophus/so3.hpp:750] SOPHUS_ENSURE failed:
SO3::exp failed! omega: nan nan nan, real: nan, img: nan
```

It fires on the very first frame, too: `last_delta_` is identity there, but `(NaN - 1.0) * 0.0` is still NaN. And `deskew` is on by default in `python/kiss_icp/config/config.py`, `config/basic.yaml`, `config/advanced.yaml` and `ros/config/config.yaml`, so this is not an opt-in path.

The most likely real-world trigger is a `PointCloud2` that declares `t` / `timestamp` / `time` / `time_stamp` but leaves it zero-filled: `GetTimestampField` keys on the field *name* only and never inspects the values. This repo can build exactly such a cloud itself — `CreatePointCloud2Msg(..., timestamp = true)` adds a FLOAT64 `time` field that `FillPointCloud2XYZ` never writes.

Other paths that reach the same place: the `rosbag` / `mcap` dataloaders hand raw timestamps straight to `Preprocess` with no normalization, so a `float32` field carrying absolute epoch seconds collapses an entire sweep to one value (one ULP of `float32` at 1.7e9 is 128 s); and per-firing-column timestamps (`ouster.py` tiles one value per column, by construction) degenerate whenever a CropBox or ROI filter narrows the cloud to a single azimuth. A single-point scan is degenerate for the same reason.

## Why the tolerance is relative

`eps * max(|min_time|, |max_time|)` rather than an absolute constant: the Python path passes raw dataset timestamps through, so the unit may be epoch seconds, nanoseconds, or an already-normalized `[0, 1]`. Scaling by magnitude is the only unit-agnostic choice, and it stays orders of magnitude below any real sweep — `eps * 1.7e9` is about 0.4 µs against a 100 ms scan. `std::abs` is load-bearing: without it a scan whose timestamps are all equal and negative (relative timing centred on zero) produces a negative tolerance and falls straight back into the division.

The comparison is written negated, `!(time_range > tolerance)`, because every comparison against NaN is false — `<=` would let a non-finite range slip past the guard and into `exp`. The guard can also manufacture such a range on its own: all-infinite timestamps give `inf - inf = NaN`. This does not sanitize an interior NaN (`{0, NaN, 1}` yields a valid range of 1, the guard correctly passes, and that one point still aborts); full timestamp sanitization is a separate concern.

## Why both files

`NormalizeTimestamps` produces the NaN *before* `Preprocess` ever runs, and `NaN <= tolerance` is false, so the core guard alone does not cover the ROS path. Filling with `1.0` maps every point to the end of the sweep, which is the identity deskew the core guard also returns, so the two halves agree and nothing is applied twice.

## Tests

`python/tests/test_preprocess.py` covers all-equal timestamps (`0.0`, `1.0`, epoch seconds), a one-ULP spread, a single-point scan, and a positive control asserting a real 0.1 s sweep is still deskewed. `python.yml` already builds the pybind extension and runs `pytest ./python/` on seven OS/arch combinations, so this needs no CI changes.

Against the unpatched code these tests do not fail cleanly — they kill the interpreter, because the abort happens in C++ inside `tbb::parallel_for`:

```
python/tests/test_preprocess.py Fatal Python error: Aborted
  File ".../kiss_icp/preprocess.py", line 46 in preprocess
  File ".../tests/test_preprocess.py", line 51 in test_deskew_is_a_noop_when_all_timestamps_are_equal
PYTEST EXIT=134
```

CI still goes red and names the test. The `ros/src/Utils.hpp` half gets no automated coverage — `ros.yml` runs `colcon build` only, with no `colcon test`.

### DIFF
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -32,7 +32,9 @@
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 #include <algorithm>
+#include <cmath>
 #include <functional>
+#include <limits>
 #include <vector>
 
 namespace kiss_icp {
@@ -62,12 +64,13 @@ std::vector<Eigen::Vector3d> Preprocessor::Preprocess(const std::vector<Eigen::V
             const auto &[min, max] = std::minmax_element(timestamps.cbegin(), timestamps.cend());
             const double min_time = *min;
             const double max_time = *max;
-            if (max_time == min_time) {
+            const double time_range = max_time - min_time;
+            const double tolerance = std::numeric_limits<double>::epsilon() *
+                                     std::max(std::abs(min_time), std::abs(max_time));
+            if (time_range <= tolerance) {
                 return frame;
             }
-            const auto normalize = [&](const double t) {
-                return (t - min_time) / (max_time - min_time);
-            };
+            const auto normalize = [&](const double t) { return (t - min_time) / time_range; };
             const auto &omega = relative_motion.log();
             std::vector<Eigen::Vector3d> deskewed_frame(frame.size());
             tbb::parallel_for(
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -67,7 +67,8 @@ std::vector<Eigen::Vector3d> Preprocessor::Preprocess(const std::vector<Eigen::V
             const double time_range = max_time - min_time;
             const double tolerance = std::numeric_limits<double>::epsilon() *
                                      std::max(std::abs(min_time), std::abs(max_time));
-            if (time_range <= tolerance) {
+            // Comparison is negated on purpose: a non-finite range must also skip deskewing
+            if (!(time_range > tolerance)) {
                 return frame;
             }
             const auto normalize = [&](const double t) { return (t - min_time) / time_range; };
--- a/cpp/kiss_icp/core/Preprocessing.cpp
+++ b/cpp/kiss_icp/core/Preprocessing.cpp
@@ -62,6 +62,9 @@ std::vector<Eigen::Vector3d> Preprocessor::Preprocess(const std::vector<Eigen::V
             const auto &[min, max] = std::minmax_element(timestamps.cbegin(), timestamps.cend());
             const double min_time = *min;
             const double max_time = *max;
+            if (max_time == min_time) {
+                return frame;
+            }
             const auto normalize = [&](const double t) {
                 return (t - min_time) / (max_time - min_time);
             };
--- a/python/tests/test_preprocess.py
+++ b/python/tests/test_preprocess.py
@@ -1,0 +1,79 @@
+# MIT License
+#
+# Copyright (c) 2022 Ignacio Vizzo, Tiziano Guadagnino, Benedikt Mersch, Cyrill
+# Stachniss.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+import numpy as np
+import pytest
+
+from kiss_icp.preprocess import Preprocessor
+
+# Points far enough from the origin to survive the range filter before and after deskewing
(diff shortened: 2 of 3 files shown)


fallenmi (changes requested):
Your note about the two-ULP boundary is a real correctness blocker, and the same tolerance has a second failure at zero.

At exact head `4a4f7fa4f8ca6980fce52f79be70819ccce25e6f`, I reproduced both boundaries against `Preprocessor`:

- timestamps at `1.0` separated by two ULP (`4.440892098500626e-16` timestamp units) pass `time_range > eps * max(abs(min), abs(max))`; normalization stretches that representation noise across `[0, 1]`, producing a full 1.0 m maximum point displacement for a 1 m relative translation;
- timestamps `0.0` and `nextafter(0.0, +inf)` are only one ULP apart (`4.940656458412465e-324`), but `eps * scale` underflows to zero, so the head again applies a full 1.0 m correction.

The new one-ULP test covers origins `1.0` and `1.7e9`, but not zero, and there is no two-ULP boundary test. Please make the threshold denormal-safe and leave several ULPs of headroom in both `Preprocessing.cpp` and `Utils.hpp`, then add regressions for these boundaries. As one verified core-side option, a local `4 * max(denorm_min, eps * scale)` experiment made both red tests pass, preserved deskewing for a real 100 ms sweep, and kept the full Python suite at 8/8.

For verification, exact base `1ffa7d7512f10bfc8b1185095011fa31184019e3` aborts on equal timestamps with exit 134. Exact head's focused suite passes 7/7 and the full Python suite passes 8/8, while each added boundary regression fails with a 1.0 m maximum difference. I built locally on macOS arm64/Python 3.11 with `scikit-build-core==0.7.1`, because the public job currently resolves unpinned 1.0.3 and stops before project build.

AI disclosure: I used OpenAI Codex to assist with this review; I reproduced the issue locally against exact head `4a4f7fa4f8ca6980fce52f79be70819ccce25e6f`.

gistrec:
Both boundaries reproduce here, and both are fixed in a40416a.

I rebuilt the pybind extension at each step on macOS arm64 / Python 3.11 rather than reasoning about the arithmetic, because the two cases differ in kind and I wanted the compiled behaviour. Against exact head `4a4f7fa`, driving `Preprocessor.preprocess` with a 1 m forward motion guess:

| timestamps | time range | head `4a4f7fa` | `a40416a` |
| --- | --- | --- | --- |
| equal, and 1 ULP at `1.0` / `1.7e9` | `0` / 1 ULP | skip | skip |
| 2 ULP at `1.0` | `4.441e-16` | **deskew, 1.0 m** | skip |
| 2 ULP at `1.7e9` | `4.768e-07` | **deskew, 1.0 m** | skip |
| `0.0` → smallest subnormal | `4.941e-324` | **deskew, 1.0 m** | skip |
| real 100 ms sweep at `1.7e9` | `1.0e-01` | deskew | deskew |

So the ULP case is what I raised above and you are right that it should not ship as a judgement call, and the underflow at zero is a genuine second bug I had missed. Thanks for catching it.

**On the threshold.** I went with a floor at the smallest *normal* double rather than a multiple of `denorm_min`:

```cpp
const double scale = std::max(std::abs(min_time), std::abs(max_time));
const double tolerance = std::max(std::numeric_limits<double>::min(),
                                  4.0 * std::numeric_limits<double>::epsilon() * scale);
```

I did build your `4 * max(denorm_min, eps * scale)` variant first, and it closes the single-subnormal case but not the family. With `min_time = 0.0` and `max_time = 8 * denorm_min`, `eps * scale` still underflows, so the threshold is pinned at `4 * denorm_min ≈ 1.98e-323` while the range is `3.95e-323` — it passes the guard and applies the full 1.0 m correction again. A floor of `2.2e-308` covers the whole subnormal region in one go, and it is still some 300 orders of magnitude below any timestamp a sensor can produce, so it cannot suppress a real sweep. Four ULPs works out to 1.5 µs at epoch-second magnitudes, three orders below the shortest real scan.

The same threshold is applied in `ros/src/Utils.hpp`.

**Regressions.** `test_deskew_is_a_noop_when_timestamps_differ_by_a_few_ulps` is now parametrized over 1, 2 and 4 ULPs at both origins, and `test_deskew_is_a_noop_for_a_subnormal_time_range_at_the_origin` covers 1 and 8 subnormals at zero. Reverting only the C++ change leaves those six red and the eight pre-existing cases green. Worth noting that these fail cleanly with an assertion rather than aborting the interpreter the way the original tests do against unpatched code — the existing guard already keeps the NaN away from Sophus, so what is left is a wrong value, not a crash.

**Two things this still does not cover,** which I would rather keep out of this PR than half-solve:

- an interior NaN, e.g. `{0.0, NaN, 0.5, 1.0}`, produces a valid range of 1.0, passes the guard correctly, and still aborts in `SO3::exp`;
- a single corrupt stamp, e.g. `{1.7e9, 1.7e9, 1.7e9, 1e300}`, gives a range so large that no relative tolerance can reject it; the three real points normalize to ~0 and take the full correction.

Both are per-point timestamp sanitization rather than a degenerate range, and both want a different fix — rejecting non-finite stamps up front, or clamping to a physically plausible sweep length. Happy to open a follow-up if you want it.

**Unrelated, but it blocks verification of any of this:** `python_package` currently fails on every OS in about nine seconds, on `main` as well as here, with

```
ERROR: Use cmake.version instead of cmake.minimum-version with scikit-build-core >= 0.8
```

`cmake.minimum-version` in `python/pyproject.toml` has been removed upstream, so the job stops before it builds the project and `pytest ./python/` never runs. I had to pin `scikit-build-core==0.7.1` locally to build at all. That means the tests in this PR are not actually being exercised by CI today. It is a one-line change and I am happy to send it as a separate PR if that is useful.

